PR #14, opened 2021-09-29
fix(assets): the lagecy behaviour, only images belong in assets/

joanise:
In 0.1.1, 0.1.0, 0.0.7 and older versions, only images got `assets/` added in front of the path provided. The smil, xml and audio files were loaded where specified. If they were in `assets/`, the paths in the `index.html` had to explicitly say that.

Thus, revert the default behaviour, i.e., the `useAssetsFolder=True` behaviour, to look for images in `assets/`, but not files directly linked from the html page.


roedoejet (approved):
This makes sense to me! Thanks to both @deltork and @joanise !

PR #27, opened 2022-09-07
Added tooltips for UI's buttons.

SamuelLarkin:
For user's convenience and to provide them with a slightly better UI experience, I propose to add tooltips for all 6 UI buttons.  Note that the buttons already have the `@aria-label` but this isn't the same thing as `@title`.  
```
Accessible Rich Internet Applications (ARIA) defines ways to make Web content and Web applications (especially those developed with Ajax and JavaScript) more accessible to people with disabilities. For example, ARIA enables accessible navigation landmarks, JavaScript widgets, form hints and error messages, live content updates, and more.
```
With this in mind, I think it is still justified to add `@title` to the UI's buttons.

[What is the difference between aria-label and title attributes?](https://stackoverflow.com/questions/27953425/what-is-the-difference-between-aria-label-and-title-attributes)

### DIFF
--- a/src/components/read-along-component/read-along.tsx
+++ b/src/components/read-along-component/read-along.tsx
@@ -1014,6 +1014,7 @@ export class ReadAlongComponent {
    */
 
   PlayControl = (): Element => <button data-cy="play-button" disabled={!this.isLoaded} aria-label="Play"
+                                       title="Play audio recording"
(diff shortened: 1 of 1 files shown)


joanise (approved):
Looks good to me. Simple and effective change.

PR #59, opened 2023-01-06
cleanup chore: created a separate file for type definitions

deltork:
moved the type definitions into a separate file for easy update and reuse


packages/stencil-component/src/components.d.ts
```diff
@@ -4,18 +4,19 @@
  * This is an autogenerated file created by the Stencil compiler.
  * It contains typing information for all components that exist in this project.
  */
-import { HTMLStencilElement, JSXBase } from "@stencil/core/internal";
-import { InterfaceLanguage, ReadAlongMode } from "./components/read-along-component/read-along";
+import {HTMLStencilElement, JSXBase} from "@stencil/core/internal";
+import {InterfaceLanguage, ReadAlongMode} from "./index.ds";
```
joanise, line 8: @deltork I notice you've changed the spacing in the import statements. In Studio-Web, we're using `npx pretty-quick` to auto-format commits, and it would undo this change. I'm not attached to any particular convention for JS yet, but I'd prefer consistency if possible. Do you use a different automated formatter or were these manual edits? 

I'd actually be inclined to configure Web-Component to use `npx pretty-quick` as well, and setup Husky, and commitlint too, while I'm at it, to align this repo with our other ones.

It would be a big change if I enabled it, affecting about a thousand lines in the repo, though, so I'm not doing it without first getting consensus!

joanise (commented):
The refactoring is a good idea, thanks for doing that!

packages/stencil-component/src/components.d.ts
joanise: Is prettier.el consistent with pretty-quick?

deltork: that file is auto generated by stencil when you run the build command

deltork: However it is good idea to use something like pretty-quick instead of using my IDE built in pretty format

joanise: Oh, I see.

PR #60, opened 2023-01-16
Add Angular integration and demo

roedoejet:
This PR uses [Nx](https://nx.dev/) to refactor our code to an Nx monorepo and includes a lightweight Angular wrapper around the stencil library. This also opens us up to React and Vue integrations by using [the same method](https://ionic.io/blog/building-react-and-angular-component-libraries-with-stencil-and-nx). 

There are a few outstanding issues:

- To Discuss
    - possibly merge the git history of studio web here and combine in the same monorepo
- YAGNI
    - ~~the selector in ngx-web-component is still just `<read-along>` - we can manually change this, but it doesn't seem possible to get Stencil to automatically prefix the selector of components (ie. `<ngx-read-along>` instead).~~
    - ~~there are no vue and react integrations yet~~
- Help Please
    - the Cypress tests do not work
    - for some reason `nx build web-component` creates a `ngx-web-component/src/generated/proxies.ts` file that imports the components from `import { Components } from '@readalongs/web-component';` but I could only get it to work by importing from `import { Components } from '@readalongs/web-component/loader';`
    - in the web-component types build directory (`dist/packages/web-component/dist/types/`) there is a strange error where  a file and folders for the path `Users/pinea/core_web/packages/web-component/.stencil/packages/web-component/stencil.config.d.ts` is being added. Obviously this is errant, but I can't figure out why stencil is adding this. 

Also, I haven't tested publishing this to npm, but I will do that once we discuss/sort out the above.


dhdaines:
Cool!!!

> for some reason nx build web-component creates a ngx-web-component/src/generated/proxies.ts file that imports the components from import { Components } from '@readalongs/web-component'; but I could only get it to work by importing from import { Components } from '@readalongs/web-component/loader'

Probably you can fix this in the exports declaration in `package.json`?

lerna-debug.log
```diff
@@ -0,0 +1,15 @@
+0 silly argv {
```
joanise, line 1: Shouldn't this file have been gitignored?

package.json
```diff
@@ -2,7 +2,54 @@
...
+    "typescript": "^4.9.4"
+  },
+  "workspaces": [
+    "packages/*"
+  ],
+  "dependencies": {
+    "@angular/animations": "~15.1.0",
+    "@angular/common": "~15.1.0",
+    "@angular/compiler": "~15.1.0",
+    "@angular/core": "~15.1.0",
```
joanise, line 46: Studio-Web is Angular 14.2. Will this mean we'll have to upgrade Studio-Web to 15.1 before we can combine it here?

joanise: just based on the documentation, the jump from 14 to 15 seems pretty big...

packages/web-component/package.json
```diff
@@ -1,6 +1,6 @@
 {
-    "name": "@roedoejet/readalong",
-    "version": "0.1.6",
+    "name": "@readalongs/web-component",
+    "version": "0.0.1",
```
joanise, line 3: 0.2.1?

joanise (commented):
I've left a few minor comments, but I think we need to review this in person so I can ask a lot more questions!

packages/web-component/package.json
roedoejet: use `--dry-run`

PR #64, opened 2023-01-31
various monorepo updates

joanise:
We don't have CI working yet, but these changes are good, I'd like to merge them into `dev.monorepo`, so that the images and state-fix branches can be rebased onto this work.


joanise:
@roedoejet This is ready to review and merge.

joanise:
The unit tests fail because an actual test case fails, but they do actually run.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "packages/Studio-Web"]
 	path = packages/Studio-Web
-	url = git@github.com:ReadAlongs/Studio-Web.git
+	url = ../Studio-Web.git
```
roedoejet, line 3: good catch, thanks @joanise

docker-notes
```diff
@@ -0,0 +1,21 @@
...
+
+Locally:
+
+docker run --rm -it ubuntu
+
+Inside the image:
+
+apt-get update
+apt install curl git vim-nox less
+apt-get install libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb
```
roedoejet, line 11: any reason not to just have a Dockerfile?

nx.json
```diff
@@ -7,7 +7,15 @@
       }
     }
   },
-  "defaultBase": "master",
+  "defaultBase": "main",
```
roedoejet, line 10: Was this actually necessary in the end?

packages/ngx-web-component/package.json
```diff
@@ -16,12 +16,12 @@
     },
     "private": false,
     "nx": {
-        "implicitDependencies": ["@readalongs/web-component"]
+        "implicitDependencies": ["web-component"]
```
roedoejet, line 19: ARGGGGG

packages/web-component/src/index.ds.ts
```diff
@@ -0,0 +1,21 @@
+export interface Page {
```
roedoejet, line 1: Why are these moved to this location?

roedoejet (approved):
Looks good to me, thanks @joanise ! there is a lot of blood, sweat, and tears behind these (relatively) few changes. Please review my comments/questions before merging.

docker-notes
joanise: I would have to write and test it...

joanise: I mean, no. :)

nx.json
joanise: Wow, I have no recollection of this change. I don't know.

packages/ngx-web-component/package.json
joanise: Loud enough for you to hear me all the way from Ottawa!

packages/web-component/src/index.ds.ts
joanise: This is some refactoring Del was doing. I don't know all the details. It makes more sense with the rest of the page_transition PR.

PR #69, opened 2023-02-02
docs: clarify how to run the component and the app locally

joanise:


README.md
```diff
@@ -74,17 +74,48 @@ then run:
...
+Studio-Web will automatically [publish](.github/workflows/publish.yml) to https://readalong-studio.mothertongues.org/ every time there is a change to `main`. Note that you will need to have CORS enabled through an extension like [this one](https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf?hl=en) in order to have the requests between Studio-Web and the API work. You will not be able to test against the prodution ReadAlongs Studio API because of the CORS protections.
 
+#### Understanding where the components come from when you run locally
+
+When you run `nx serve Studio-Web`, that process is actually serving all the components needed
```
roedoejet, line 81: not exactly true - it's not serving them, it's just that `nx build web-component` puts the built components in the place that nx imports them (via the paths defined in the root tsconfig.base.json). I think "serve" is misleading because it makes it seem like we're putting them on a port somewhere or something.

roedoejet (changes requested):
Just small changes where my comment is - looks good though, thanks @joanise !

README.md
joanise: Subtle, but OK. I've made a patch, let me know if it is better now.

joanise:
@roedoejet I'll squash merge this PR if you are happy with it now.

PR #70, opened 2023-02-03
fix: display error when audio cannot load

joanise:
We defer the audio loading, but we still need to display an error on failure, and recognize success on successful load.


joanise:
This PR got subsumed by a subsequent one.

PR #72, opened 2023-02-06
Merge studio web into the monorepo itself

joanise:
I think this is ready to become our new main branch, as is.

Please review but do not merge: I want to do a fast-forward merge from the CLI, any other way to merge this particular PR risks creating an absolute mess of the repo.

Done:
 - replaced the Studio-Web submodule by the `packages/studio-web` subdirectory
 - merged the histories so we have a trace of both original repos - this repo now has two unrelated initial commits, I hope that doesn't get to confusing
 - CI works from the monorepo
 - commitlint activated globally
 - Aidan already had the studio-web published from the monorepo with the submodule, so that should keep working with the subdirectory
 - running tests is now uniform and well documented

Still TODO (and I think it's OK to do these after merging the PR):
 - ~~fix the publication of studio-web to npm from here instead of from the Studio-Web repo~~ (n/a, Studio-Web is deployed, not published)
 - activate pretty-quick auto formatting -- eventually -- I still think it's too early to do it because we have branches to rebase and merge and  applying pretty-quick everywhere now will make that impossible
 - the top level needs a LICENSE file and we need to figure out the correct Copyright statement to put there.

@roedoejet and @dhdaines if you can review this, that would be great, and then I'd like to forward-forward merge this into main soon so future development will continue from here.


dhdaines:
Basic build and run works well for me - we may wish to include a note to run `git clean -dfx` or to remove `packages/Studio-Web` when updating for people with case-sensitive filesystems, since testing will break otherwise.

README.md
```diff
@@ -10,7 +10,7 @@ This mono repo combines four packages:
  - A [stencil web component](packages/web-component/) for visualizing read alongs,
  - An [Angular Library](packages/ngx-web-component/) that wraps the stencil web component,
  - A [demo web application](packages/angular-demo/) to show how to use the angular library in an Angular web application.
-- The [Studio-Web](packages/Studio-Web/) application for creating ReadAlongs
+- The [studio-web](packages/studio-web/) application for creating ReadAlongs
```
dhdaines, line 13: should the name of the link still be Camel-Kebab-Cased?

packages/studio-web/.github/workflows/codeql.yml
```diff
@@ -0,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
```
dhdaines, line 1: Do we still want to have the GH Actions workflows here?

dhdaines:
getting a lot of these when running `nx test:once web-component`:

```
     AssertionError: Timed out retrying after 4000ms: Expected the subject to host a shadow root, but never found it.
https://on.cypress.io/shadow
```

packages/studio-web/.github/workflows/codeql.yml
joanise: right, no. good catch.

README.md
joanise: Right, yes, it should. I caught (and kept) some of them, but not this one.

joanise:
> getting a lot of these when running `nx test:once web-component`:
> 
> ```
>      AssertionError: Timed out retrying after 4000ms: Expected the subject to host a shadow root, but never found it.
> https://on.cypress.io/shadow
> ```

I get those errors if I don't first launch `nx serve web-component` in a separate window. 

Unfortunately, no, I still don't have a single-call command to test web-component... (Well, we do, there's a `"test"` script in `packages/web-component/package.json`, but it's broken at the moment.)

dhdaines:
> I get those errors if I don't first launch `nx serve web-component` in a separate window.

Ah, okay, yes, I have to serve the test data and the web-component (and remember to `killall node` before so it gets the right port number).  Now it works.

dhdaines (approved):
This works at least as well as it did before, I approve

roedoejet (approved):
Awesome @joanise !

PR #76, opened 2023-02-07
Use the new readalong XML format

dhdaines:
Move to the new `<read-along>` format, as produced by https://github.com/ReadAlongs/Studio/pull/142

This means that we now add the alignments to the XML internally using the DOM rather than reading an external SMIL file.  The DOM manipulation code has been simplified somewhat, and we pass around a `Document` rather than a base64 string or the XML text, in order to reduce unnecessary serialization.

Note that there is one interesting possibility which is not included in this PR but could be in the future - instead of putting the XML in a separate file, we can simply insert it as a child element of `<read-along>` since we are navigating it entirely with CSS selectors.

TODO:
- get the right version of the web component JS code in the downloaded bundle
- any missing documentation updates


Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+RUN apt-get update && apt-get -y install curl xz-utils git
+WORKDIR /root
+RUN curl https://nodejs.org/dist/v18.13.0/node-v18.13.0-linux-x64.tar.xz | tar Jxf -
+RUN mkdir /root/Web-Component
+COPY .git/ /root/Web-Component/.git/
+WORKDIR /root/Web-Component
+RUN git checkout dev.monorepo && git restore .
+RUN git submodule update --init
+RUN PATH=/root/node-v18.13.0-linux-x64/bin:$PATH npm install --force
```
joanise: This `--force` should no longer be required.

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu:latest
+
+RUN apt-get update && apt-get -y install curl xz-utils git
+WORKDIR /root
+RUN curl https://nodejs.org/dist/v18.13.0/node-v18.13.0-linux-x64.tar.xz | tar Jxf -
+RUN mkdir /root/Web-Component
+COPY .git/ /root/Web-Component/.git/
+WORKDIR /root/Web-Component
+RUN git checkout dev.monorepo && git restore .
```
joanise: Smells like a stale dockerfile, checking out out dev.monorepo. Is there a way to use the current active commit in the sandbox we're building from instead?

This dockerfile isn't actively used, I know, but if we're going to include it, I guess it should default to `main` and we can edit by hand if needed.

packages/angular-demo/src/assets/ej-fra.xml
```diff
@@ -1,28 +1,28 @@
...
     <text xml:lang="fra" id="t0">
         <body id="t0b0">
             <div type="page" id="t0b0d0">
-            <graphic url="avatar.png"/>
+            <graphic url="http://localhost:5000/ej-fra/avatar.png" id="t0b0d0graphic0"/>
                 <p id="t0b0d0p0">
-                    <s id="t0b0d0p0s0"><w id="t0b0d0p0s0w0">Bonjour</w>.</s>
-                    <s id="t0b0d0p0s1"><w id="t0b0d0p0s1w0">Je</w> <w id="t0b0d0p0s1w1">m</w>'<w id="t0b0d0p0s1w2">appelle</w> <w id="t0b0d0p0s1w3">Éric</w> <w id="t0b0d0p0s1w4">Joanis</w>.</s>
-                    <s id="t0b0d0p0s2"><w id="t0b0d0p0s2w0">Je</w> <w id="t0b0d0p0s2w1">suis</w> <w id="t0b0d0p0s2w2">programmeur</w> <w id="t0b0d0p0s2w3">au</w> <w id="t0b0d0p0s2w4">sein</w> <w id="t0b0d0p0s2w5">de</w> <w id="t0b0d0p0s2w6">l</w>'<w id="t0b0d0p0s2w7">équipe</w> <w id="t0b0d0p0s2w8">des</w> <w id="t0b0d0p0s2w9">technologies</w> <w id="t0b0d0p0s2w10">pour</w> <w id="t0b0d0p0s2w11">les</w> <w id="t0b0d0p0s2w12">langues</w> <w id="t0b0d0p0s2w13">autochtones</w> <w id="t0b0d0p0s2w14">au</w> <w id="t0b0d0p0s2w15">CNRC</w>.</s>
+                    <s id="t0b0d0p0s0"><w id="t0b0d0p0s0w0" time="0.455" dur="1.165">Bonjour</w>.</s>
```
joanise: Just curious: did you write a conversion utility from .smil + .xml to RAS format? Or did you regenerate this demo file independently?

That does raise the question, really: do we want to have such a conversion utility?

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -21,7 +21,7 @@ export class DemoComponent implements OnInit {
     pageTitle: $localize`PageTitle`,
   };
 
-  constructor(public titleService: Title, private b64Service: B64Service) {
+  constructor(public titleService: Title, public b64Service: B64Service) {
```
joanise, line 24: Why make `b64Service` public here? Shouldn't both of these services be private?

packages/studio-web/src/app/soundswallower.service.ts
```diff
@@ -101,3 +101,23 @@ export class SoundswallowerService {
     });
   }
 }
+
+export function createAlignedXML(
```
joanise, line 103: Nice idea to move this logic out to the ss service, I like it.

packages/studio-web/src/app/soundswallower.service.ts
```diff
@@ -101,3 +101,23 @@ export class SoundswallowerService {
...
+  xmlText: string,
```
joanise, line 104: So I assume this is the string returned by the `/assemble` endpoint?

packages/web-component/src/utils/utils.ts
```diff
@@ -60,60 +15,56 @@ export function zip(arrays): Array<any[]> {
...
-  let audio = zip([audio_begin, audio_duration])
-  let result = {}
-  for (var i = 0; i < text.length; i++) {
-    result[text[i]] = audio[i]
+export function extractAlignment(parsed_text: Array<Page>): Alignment {
+  let alignment = {};
+  for (const page of parsed_text) {
+    for (const p of page.paragraphs) {
+      const words = Array.from(p.getElementsByTagName("w"));
+      for (const w of words) {
```
joanise: Does this code assume that the structure is of fixed depth, with `<p>` elements always immediately under the page `<div>` elements, w under p, and no variations? Or will this search for any p elements under the page, any w elements under the paragraph, etc? It seems to me the original code was quite flexible, allowing for structures that were not necessarily exactly what Studio produces.

I don't know if we have use cases for such structures at the moment, but I could imagine wanting to use spans for changing colour or other font aspects for several words/paragraphs in a row. I don't have a concrete use case for such formatting, but if we decide we stop supporting it, I'd like to make sure we're intentionally making that decision, not accidentally.

joanise: I guess that same question/comment applies to the DTD as well.

I think I like making the structure stricter, because it will make things simpler. But again only if we're doing so deliberately.

joanise (commented):
I like this. I've left several comments in the code.

We should deliberately decide if we're choosing to narrow down the supported format, but otherwise this is good.

As we've already discussed this PR has to be merge with its sibling Studio PR, with publication to npm and unpkg all at about the same time, but I think we should proceed with all this soon.

Dockerfile
dhdaines: yeah, I am not so sure about the dockerfile in general, but I'll update it to not checkout a specific branch or do --force

packages/angular-demo/src/assets/ej-fra.xml
dhdaines: I just ran the existing .xml file through `readalong align` :)

packages/studio-web/src/app/demo/demo.component.ts
dhdaines: it has to be public because I use it in the template - the base64-ification is deferred because we are passing around a `Document` up to the last minute in order to update the images inside it on download

note that if we put the RAS XML inside the `<read-along>` element we wouldn't have this problem as it would just be there in the DOM...

packages/studio-web/src/app/soundswallower.service.ts
dhdaines: yes exactly, we parse it, add the alignments, then return the resulting `Document`

packages/web-component/src/utils/utils.ts
dhdaines: Good question, but yes, `getElementsByTagName` will find `<w>` at any depth.  It is like a CSS selector (in fact we could use `querySelectorAll` to be more explicit about this)

packages/angular-demo/src/assets/ej-fra.xml
joanise: Smart!

packages/studio-web/src/app/demo/demo.component.ts
joanise: OK, thanks for the clarification. This scoping for service injection still confuses me, I'll eventually get the hang of it.

packages/web-component/src/utils/utils.ts
joanise: Ah, OK, then all is good.

packages/studio-web/src/app/app.component.ts
```diff
@@ -120,18 +121,11 @@ export class AppComponent {
...
         this.fileService.readFileAsData$(event[1]),
-        of(
-          `data:application/xml;base64,${this.b64Service.xmlStringToB64(
-            event[2]
-          )}`
-        ),
-        of(
-          this.b64Service.alignmentToSmil(event[3] as Segment, "test", "test")
-        ),
+        of(aligned_xml),
```
roedoejet, line 127: so much nicer 🥲

packages/studio-web/src/app/b64.service.ts
```diff
@@ -11,16 +11,6 @@ import { Segment } from "soundswallower";
   providedIn: "root",
 })
 export class B64Service {
-  smilTemplate = `<smil xmlns="http://www.w3.org/ns/SMIL" version="3.0">
```
roedoejet, line 14: I didn't fully realize how much nicer this makes the serialization stuff work when we don't have two separate files like this. You're a wizard @dhdaines 🥲

packages/web-component/test-data/ej-fra/index.html
```diff
@@ -9,11 +9,12 @@
...
     <body>
         <!-- Here is how you declare the Web Component. Supported languages: en, fr -->
-        <read-along use-assets-folder="false" text="ej-fra.xml" alignment="ej-fra.smil" audio="ej-fra.m4a" language="fra">
+        <read-along use-assets-folder="false" readalong="ej-fra.xml" audio="ej-fra.m4a" language="fra">
             <span slot="read-along-header">Insert Title Here Too</span>
         </read-along>
     </body>
 
     <!-- The last step needed is to import the package -->
+    <!-- Note that you will probably have to change 3333 to whatever port `npm serve-test-data` ends up listening on, sorry. -->
```
roedoejet: but serve-test-data doesn't serve the web-component right? these notes shouldn't be here I don't think...and in any case it's `nx serve-test-data web-component` now ;)

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -747,9 +740,8 @@ export class ReadAlongComponent {
    * is being read
    */
   componentDidLoad() {
-    this.processed_alignment = parseSMIL(this.alignment)
-    this.assetsStatus.SMIL = Object.keys(this.processed_alignment).length ? LOADED : ERROR_LOADING
-
+    this.processed_alignment = extractAlignment(this.parsed_text)
+    this.alignment_failed = (Object.keys(this.processed_alignment).length == 0)
```
roedoejet, line 744: @joanise - there are going to be conflicts here with Del's error handling.

roedoejet (approved):
These are a lot of great changes to the codebase, bravo @dhdaines 👏 

I still feel like the question from https://github.com/ReadAlongs/Studio/pull/142 of whether to use .ras is open and is obviously relevant to this PR, and ideally I would see some versioning code in here too. I think this can wait for a subsequent PR though.

I have a minor comment around clarifying the language in the test-data examples around fixing it for a specific port.

packages/web-component/test-data/ej-fra/index.html
dhdaines: ahaha! yes, that should say `nx serve web-component` in fact

packages/web-component/src/components/read-along-component/read-along.tsx
dhdaines: yes ... lots of them, particularly since I had to reformat the TSX a bit to be able to read it ... I will try to find a solution once this is merged

packages/web-component/test-data/ej-fra/index.html
joanise: Yeah, this is a left-over note from when I still didn't really understand what I was doing here. That line is repeated in all the test files! We can remove it, but it's not germane to this PR.

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: Yeah, This PR will conflict with that PR is many ways...

PR #77, opened 2023-02-07
Compress images reasonably by default

dhdaines:
Nothing fancy, just JPEG at 0.75, which should be good enough for most cases to avoid absurdly large output files.


joanise (approved):
Gotta love it when stuff is that easy! Wish it was all like that.

roedoejet (approved):
hah! so easy! nicely done

PR #62, opened 2023-01-23
Patch.page transition

deltork:

PR #79, opened 2023-02-07
multiple features: page transition, async loading, better errors, etc

joanise:
This PR is meant to replace #62 
It has been rebased on the monorepo and can merge cleanly at the moment.

It will still run into conflicts with #76, unfortunately.

### DIFF
--- a/packages/web-component/cypress/e2e/basic.cy.js
+++ b/packages/web-component/cypress/e2e/basic.cy.js
@@ -8,6 +8,7 @@ context("The Readalong Component", () => {
 
   const FOR_PAGE_TURN_ANIMATION = 500; // ms
   const FOR_ERIC_TO_TALK_A_BIT = 3000; // ms
+  const FOR_ERIC_TO_TALK_A_BIT_LESS = 1500; // ms
 
   beforeEach(() => {
     cy.intercept(/\.js/).as('page')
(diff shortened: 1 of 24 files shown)


joanise:
Outstanding issues:
 - loop { play, maybe pause, click around } will still eventually get the player in a bad state where you can't remind playing anymore
 - In http://localhost:5000/ej-fra/index-missing-alignment.html there is an error, so the play/pause/gear buttons should be inactive, but they are not. If you click the gear menu, you get an empty pane, and an exception in the console
 - testing passes on my computer but not in CI.
 - When I load http://localhost:5000/ej-fra/index-malformed-xml.html , the gear menu appears over the main window and glides away. It really should just never get rendered until the gear button gets pressed.

joanise:
CI: currently, most tests pass, but not all.
See, for example, https://github.com/ReadAlongs/Web-Component/actions/runs/4116395685/jobs/7106393731
 - 17 tests pass
 - 4 tests fail, with failure to load here, uncaught exception there, etc. I cannot reproduce any of this locally...

dhdaines:
You can merge this one first once it's ready, then I will handle the merging of #76 myself

joanise:
> You can merge this one first once it's ready, then I will handle the merging of #76 myself

Great, thank you! I was dreading that merge. It's not ready yet, though, I'd like CI to pass at least, and ideally fix the click/play bug.

joanise:
These changes have been redone in different PRs.

PR #85, opened 2023-02-10
feat: add editable translations

roedoejet:
added to both web component and studio-web


packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1014,40 +1057,59 @@ export class ReadAlongComponent {
    *
    * A sentence element with one or more words
    */
-  Sentence = (props: { words: Node[], attributes: NamedNodeMap }): Element => {
-    if (!this.hasTextTranslations && props.attributes["class"]) {
-      this.hasTextTranslations = props.attributes["class"].value.match("translation") != null;
+  Sentence = (props: { sentenceData: any | null }): Element => {
```
joanise: Can't we have something more precise than `any` here?

packages/web-component/src/scss/_shame.scss
```diff
@@ -1,3 +1,11 @@
+[contenteditable][data-placeholder]:empty:before {
```
joanise, line 1: That's funny, putting shameful code in `_shame.scss`! 
Is this based on the idea here https://csswizardry.com/2013/04/shame-css/ ?

joanise: and since you brought my attention here, I have to ask, are any of the hacks in this file fixable by now?

packages/web-component/test-data/ej-fra/index-edit.html
```diff
@@ -0,0 +1,21 @@
...
+<head>
+    <meta charset="UTF-8">
+    <title>Insert Title Here</title>
+    <!-- Import fonts. Material Icons are needed by the web component -->
+    <link href="https://fonts.googleapis.com/css?family=Lato%7CMaterial+Icons%7CMaterial+Icons+Outlined" rel="stylesheet">
+</head>
+
+<body>
+    <!-- Here is how you declare the Web Component. Supported languages: en, fr -->
+    <read-along mode="EDIT" use-assets-folder="false" href="ej-fra.ras" audio="ej-fra.m4a" language="fra">
```
joanise, line 13: Is `use-assets-folder="false"` required for image editing to work properly now?

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -451,6 +452,14 @@ export class ReadAlongComponent {
     return this.images
   }
 
+  /**
+  * Get Translations
+  */
+  @Method()
+  async getTranslations(): Promise<object> {
```
dhdaines, line 479: Any particular reason why this is async?

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1014,40 +1057,59 @@ export class ReadAlongComponent {
    *
    * A sentence element with one or more words
    */
-  Sentence = (props: { words: Node[], attributes: NamedNodeMap }): Element => {
-    if (!this.hasTextTranslations && props.attributes["class"]) {
-      this.hasTextTranslations = props.attributes["class"].value.match("translation") != null;
+  Sentence = (props: { sentenceData: any | null }): Element => {
+    let words: ChildNode[] = Array.from(props.sentenceData.childNodes)
+    let attributes: NamedNodeMap = props.sentenceData.attributes
```
dhdaines: Suggest using `hasAttribute()` and `getAttribute()` instead of this as they have more precise semantics for existing or non-existing attributes

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1014,40 +1057,59 @@ export class ReadAlongComponent {
...
+              </span>
+          } else {
+            return <button data-cy="add-translation-button" class="sentence__translation" onClick={() => this.addLine(props.sentenceData)}>{this.returnTranslation('add-line', this.language)}</button>
+          }
+        }
+       else {
+          return null
+       }
+   })()
+   }
```
joanise, line 1141: These lines are not indented correctly, and that made it harder to understand the code. We really need to enable prettier on this repo!

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -35,6 +35,29 @@ export class DemoComponent implements OnInit {
...
   ngOnInit(): void {}
 
+  async updateTranslations(doc: Document): Promise<boolean> {
+    const translations: any = await this.readalong.getTranslations();
+    if (Object.keys(translations).length == 0) {
+      return false;
+    } else {
+      const sentence_nodes = doc.querySelectorAll("s")
+      // represents all translation nodes that have already been added
+      const translation_node_ids = Array.from(doc.querySelectorAll(".sentence__translation")).map((t_node) => t_node.id)
```
dhdaines: Suggest using `Set` here instead if it isn't a problem to polyfill it (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -106,6 +106,7 @@ export class ReadAlongComponent {
   current_page;
   hasTextTranslations: boolean = false;
   @State() images: object;
+  @State() translations: object;
```
dhdaines: Could use an index signature here (... and for `images` too while we're at it)

packages/web-component/src/components/read-along-component/read-along.tsx
roedoejet: > Any particular reason why this is async?

https://stenciljs.com/docs/methods#public-methods-must-be-async

dhdaines: oh, I see! :)

dhdaines (changes requested):
Looks good, thanks! Just a couple of suggestions which are not essential...

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -845,6 +854,18 @@ export class ReadAlongComponent {
       "loading": {
         "eng": "Loading...",
         "fra": "Chargement en cours"
+      },
+      "add-line": {
+        "eng": "Add line",
```
joanise: I would prefer to use "translation" rather than "line", like "Add translation" and "Remove translation".

I guess you didn't do that because this can also be used for transliteration or whatnot, but just "Add line" is not intuitively clear.

packages/web-component/src/components/read-along-component/read-along.tsx
roedoejet: good point, thanks @dhdaines !

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -845,6 +854,18 @@ export class ReadAlongComponent {
...
+        "fra": "Ajouter une ligne" 
+      },
+      "remove-line": {
+        "eng": "Remove line",
```
joanise: This is the wrong place to comment on this, I know, but I found a bug: if I add a translation, click the download button to get my RA.html, then click "Remove Line", the line disappears from the edit interface but if I re-download, the deleted line still shows in the downloaded RA.html.

packages/web-component/src/components/read-along-component/read-along.tsx
roedoejet: good point too. thanks @dhdaines - I'll make these changes.

joanise (changes requested):
Nice work, done efficiently as you always do!

Issues:
 - Remove translation is buggy, the translation is not removed from the downloaded single-file RA.html under this workflow: Add line; download; Remove line; download. The second downloaded RA still has the deleted line.
 - When you click "Add line", the edit box is created, but the cursor is not in it, you have to click a second time to put it there. Could that happen automatically?
 - The line edit box lets you insert a newline, and shows the translation on two lines while you're editing it, but when you download it's just shown on one line. I'd like to be able to preserve newlines entered by the user, ideally, or else allow two "Add line" clicks for the same sentence.

packages/studio-web/src/app/demo/demo.component.ts
roedoejet: so:

```typescript
    const sentence_nodes = doc.querySelectorAll("s")
      // represents all translation nodes that have already been added
      const translation_node_ids = new Set(Array.from(doc.querySelectorAll(".sentence__translation")).map((t_node) => t_node.id))
      sentence_nodes.forEach((sentence: HTMLElement) => {
        if (sentence.id in translations && translation_node_ids.has(sentence.id)) {
          let newSentence = document.createElement('s')
          newSentence.setAttribute('do-not-align', "true")
          newSentence.setAttribute("id", sentence.id)
          newSentence.setAttribute("class", "sentence__translation")
          newSentence.setAttribute("xml:lang", "eng")
          newSentence.innerText = translations[sentence.id]
          sentence.insertAdjacentElement('afterend', newSentence)
        }
      })
```
?

dhdaines: Looks good!  A bit more readable - is the `Array.from` necessary when you create `translation_node_ids`?  Sometimes it is and sometimes it isn't, which is a pain...

packages/web-component/test-data/ej-fra/index-edit.html
roedoejet: Hm, I don't think so. I haven't tried that though

packages/web-component/src/scss/_shame.scss
roedoejet: yea pretty similar to that. I'm not sure. probably, but that kind of refactoring isn't on this sprint for me :)

packages/web-component/src/components/read-along-component/read-along.tsx
roedoejet: yea exactly, it's not necessarily a translation. I think it's intuitive what it is after you click it though no?

roedoejet:
I'm also getting a 422 when I try and submit a .ras that already has translations (like ej-fra-translated.ras):

```
file.service.ts:22          POST http://localhost:8000/api/v2/assemble 422 (Unprocessable Entity)
```

So there appears to be an issue with the Studio API

dhdaines:
> I'm also getting a 422 when I try and submit a .ras that already has translations (like ej-fra-translated.ras):
> 
> ```
> file.service.ts:22          POST http://localhost:8000/api/v2/assemble 422 (Unprocessable Entity)
> ```
> 
> So there appears to be an issue with the Studio API

Ah, hmm, I'll have to take a look at that again... it may be a problem with the DTD

packages/web-component/src/components/read-along-component/read-along.tsx
roedoejet: yup. good call. I'll change it.

roedoejet:
@joanise - I fixed the bug you mentioned (good catch!)

I haven't addressed your two other issues: the automatic placement of the cursor and the multiline discrepancy, and I won't get to those today. Signing off...

dhdaines:
> > I'm also getting a 422 when I try and submit a .ras that already has translations (like ej-fra-translated.ras):
> > ```
> > file.service.ts:22          POST http://localhost:8000/api/v2/assemble 422 (Unprocessable Entity)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So there appears to be an issue with the Studio API
> 
> Ah, hmm, I'll have to take a look at that again... it may be a problem with the DTD

ah, oops, yes. I somehow forgot to declare `class` as an attribute for `<div>` (will add it to the others as well)

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: No, I don't think so. I was playing with it with Pat, too, and neither of us likes it. "Line" is not great because that suggests it's a line in the RA itself, i.e., we're adding more aligned text, and that's because of how we use "line" in the instructions for creating the text in the first step. So "translation" is much better than "line", and something that subsumed translation, transliteration, gloss, etc would be even better, if we can come up with some such.

joanise: @roedoejet Maybe "Add tr." / "Ajouter tr." would work: "tr." can be short for both translation and transliteration, and in the documentation we can say it can also be a gloss.

joanise:
We don't get a 422 anymore when translations are already in the file (e.g., `ej-fra-translated.ras`) but the result is not great, because now we can add a new translation line underneath each existing translation, so it's not recognizing that those are already translations.
image

And then when clicked on the first two "Add line" buttons I ended up with this:
image

joanise:
Next problem that `ej-fra-translated.ras` highlights, is that we don't handle very well the case where the .ras file already had some images in it:
image
although if we click on the garbage can then we can add a new image in its place, so it's recoverable.

roedoejet:
This is ready for review again. There are a few additional changes wrapped up in this PR:

- [f60f89c](https://github.com/ReadAlongs/Web-Component/pull/85/commits/f60f89c7ff114d95432a30de69c2b10d24509d85) fixes the style bug that Del talked about around non-word text. Sorry this is bundled in the commit that fixes the other bugs you talked about. It's two lines in the scss so hopefully it doesn't distract
- There was a number of nasty bugs that came up around images - basically because of the way we were loading images (as properties on the page) it caused a series of rendering problems that conflicted with the `contentEditable` translation elements (ex. if you deleted an Image, you would lose your translations etc). I re-factored this so that images are only loaded from the `this.images` internal state variable, and while I was at it, I included the functionality we discussed about loading .ras files (it makes a HEAD request for the resource and doesn't load it into the `this.images` variable if the request doesn't return a 2XX status. Maybe this should have been in a separate PR, but there was an interplay between images and translations that was causing a bug. Hopefully it's not too hard to review together.
- All the other fixes are done, except I didn't change the "Add line". what did we decide? Plus sign emoji? "Add note"?

joanise:
> * All the other fixes are done, except I didn't change the "Add line". what did we decide? Plus sign emoji? "Add note"?

Not sure we settled on a choice in the end, but I think the consensus was towards a fat Plus sign, which I guess would mean a fat Minus sign for removing.

joanise:
There's a new bug, sorry. Much is indeed better, but...

Steps:
 - add an image
 - download 
 - delete the image
 - add a new image
 - download => the first image is still there.

Current workaround:
 - delete the image
 - download
 - add the new image
 - download => the new image is now shown
If I add an image

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -35,32 +35,63 @@ export class DemoComponent implements OnInit {
...
+      // represents all translation nodes that have already been added
+      const translation_node_ids = Array.from(doc.querySelectorAll(".editable__translation")).map((t_node) => t_node.id)
+      sentence_nodes.forEach((sentence: Element) => {
+        // Add a translation
+        if (sentence.id in translations && translation_node_ids.indexOf(sentence.id) < 0) {
+          let newSentence = document.createElement('s')
+          newSentence.setAttribute('do-not-align', "true")
+          newSentence.setAttribute("id", sentence.id)
+          newSentence.setAttribute("class", "sentence__translation editable__translation")
+          newSentence.setAttribute("xml:lang", "eng")
```
joanise, line 53: Hard-coded eng? Why?

Why add xml:lang at all?

packages/web-component/cypress/e2e/edit.cy.js
```diff
@@ -0,0 +1,33 @@
+context("The Readalong Component", () => {
+    /**
+     * Wait for the audio and the SMIL to load.
+     */
+    const EXPECTED_LOADING_TIME = 2000; // ms
```
joanise, line 5: Turns out we don't need this at all. I simply removed it in some .cy.js files, and they still work, because I think when you look for something, Cypress already gives it a changes to should up.

Del also has a different (and much more efficient) wait variant that you can see in https://github.com/ReadAlongs/Web-Component/blob/dev.hide-tr-at-load/packages/web-component/cypress/e2e/customizations.cy.js:

```
context('Testing creator enabled settings', () => {
  beforeEach(() => {
    cy.intercept(/\.ras/).as('text')
    cy.intercept(/\.m4a/).as('audio')
  });

  it('testing creator enforced hiding of translation at load time ', function () {

    cy.visit("/ej-fra/index-translated-no-display.html");
    cy.wait('@text')
    cy.wait('@audio')
    cy.readalong().within(() => {
    ...
```

This version waits exactly as long as needed and no more. Our CI runs much faster now in that branch, so we should apply the same idea to new test suites.

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -856,19 +894,38 @@ export class ReadAlongComponent {
...
-    props.pageData.img = imageURL
     let newImage = {}
     newImage[pageIndex] = imageURL
     this.images = {...this.images, ...newImage}  // Using spread operator as advised https://stenciljs.com/docs/reactive-data#updating-an-object
   }
 
-  deleteImage(props){
-    let pageIndex = this.parsed_text.map((page: Page) => page.id).indexOf(props.pageData.id)
-    delete props.pageData.img
+  deleteImage(pageIndex: number){
```
joanise, line 940: nicely simplified code.

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -900,10 +957,10 @@ export class ReadAlongComponent {
    *
    * @param props
    */
-  RemoveImg = (props: { pageData: Page }): Element => {
+  RemoveImg = (props: { pageIndex: number }): Element => {
     return (<button data-cy="delete-button" aria-label="Delete"
     title="Delete image"
-    onClick={() => this.deleteImage(props)}
+    onClick={() => this.deleteImage(props.pageIndex)}
```
joanise, line 975: much cleaner code, passing the narrow value we need instead of the big object.

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1014,40 +1077,56 @@ export class ReadAlongComponent {
...
+      class={'sentence' + " " + (props.sentenceData.hasAttribute('class') ? props.sentenceData.getAttribute('class') : "")}>
       {
         /* Here are the Word and NonWordText children */
-        props.words.map((child: Element, c) => {
+        words.map((child: Element, c) => {
 
           if (child.nodeName === '#text') {
             return <this.NonWordText text={child.textContent} attributes={child.attributes}
-              id={(props.attributes["id"] ? props.attributes["id"].value : "P") + 'text' + c} />
+              id={(props.sentenceData.hasAttribute('id') ? props.sentenceData.getAttribute('id') : "P") + 'text' + c} />
```
joanise, line 1114: Not a change request, but I wonder if this could be shortened with a short-circuit OR style construct, like

    id = {(props.sentenceData.getAttribute('id') || "P") ...

Caveat: I have no idea if that's even valid TS, or if it might throw an exception instead of doing what I expect, although https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute says `getAttribute()` returns `null` when the element does not exist, so it ought to work.

joanise (changes requested):
Much improved, but still a bug to fix with images, and the + instead of "add line".

joanise:
This PR doesn't merge cleanly anymore. I've started trying to merge main into it and resolving the conflicts, but CI has a failure in basics.cy.js for me now, with the merge. I'll see if I can fix it and put my merge, maybe in a separate branch so you can test it too.

packages/web-component/src/components/read-along-component/read-along.tsx
dhdaines: Yes, I agree, it is nearly always better to use `getAttribute` than directly looking in `attributes`.  `undefined` is one of the many awful flaws of JavaScript and should be banished from the face of the earth

dhdaines: Ah on further reflection, I disagree with using an or-construct, because what if `id` is `0` or `""` or something else un-truthy?  We should *always* explicitly test for things existing instead of relying on this, it is one of the worst flaws of JavaScript that you can access any attribute or key and just get `undefined` (at least Perl had the decency to not have different `undef` and `null` values)

joanise: Right, that makes sense. So the code should stay as it was already, since it's indeed testing explicitly for what we actually want to know.

roedoejet:
I added titles to the buttons but found this issue: https://github.com/ReadAlongs/Web-Component/issues/105

packages/web-component/cypress/e2e/edit.cy.js
joanise: Well, never mind that claim, apparently we do need to wait. Sigh.

joanise (approved):
All good now, as far as I'm concerned.

PR #95, opened 2023-02-13
Responsive image edit

### LINKED ISSUE #87: Web-C in EDIT mode: the select some files box for adding images is not responsive
When you view an RA in the Web-C in VIEW mode, the image takes only up to about half the width, even on a narrow cell phone phone.

When you edit an RA in the Web-C in EDIT mode, the "Select some files" box is fixed width, needs to be responsive too, with max 40% or 50% of width.

image

Even when you scroll right, the text is all scrunched up and there is no way to see all of it:

image

joanise:
Make the drop image box responsive to screen size, so it shrinks when the screen is smaller.

Also including a commit that renames `index.ds.ts` to `index.d.ts`, which what Del confirms that file was supposed to be called in the first place.

Fixes: #87


dhdaines (approved):
Great! I always wondered what a `.ds.ts` file was supposed to be, figured it was some Stencil magic, but indeed, it *looks* like a TypeScript definitions file, it *quacks* like a TypeScript definitions file...

PR #96, opened 2023-02-13
feat: sort the language list by the language name, not code

### LINKED ISSUE #94: Sort the list of language by language name, not code
Hi Studio-Web, it's counterintuitive to loop for the order of language codes instead of language names. We should sort them by name before displaying them.

joanise:
This is just because it's not intuitive to look for language in the drop-down list order of codes while lang names is what the display starts with.

Fixes: #94


roedoejet (approved):
Sounds good.

PR #99, opened 2023-02-13
Studio web tweaks

### LINKED ISSUE #32: When recording fails, we get both an error message and a Yay success message
If I click on record with my microphone turned off, I get this:
image
and the Play, Download and Delete buttons appear.

If I click Play, nothing happens, and if I click Download I get a `.webm` file containing just `undefined`.

The correct behaviour would be to prevent the Yay! message from being displayed and to have just the Record button displayed, maybe by setting `this.audioControl.setValue(null);` like in `deleteRecording()`?

### LINKED ISSUE #86: Studio-Web: hidden text still gets used
Steps to reproduce:
 - type text into Text in Write mode
 - click on File mode
 - click on Write mode  --> the text box is empty
 - click on Go to the next step!

Result:
 - the alignment uses the text initially written, even though the text box was empty when you clicked back on text mode

Preferable:
 - when you click on Write mode again, the text that is remembered is displayed again and can be further edited

### LINKED ISSUE #89: Studio-W: warn user to listen to their recorded audio before continuing
When I try to record, especially the first time in a given session, I often have to try multiple times before it's right. Recording doesn't actually start right away half the time, so I record but the first few words are missing.

That's going to be hard to fix, but we should warn the user: after recording, two notes should appear between "re-record" and "play/download/delete":
 - "Before continuing, play back you audio to verify it recorded correctly."
 - "If your audio is good, download it in case you need to restart the app."

or maybe something shorter to that general effect.

joanise:
- display remembered text (Fixes #86)
 - better message for some audio recording error (Fixes #89)
 - warn the user to vet recorded audio (Fixes https://github.com/ReadAlongs/Studio-Web/issues/32)

### DIFF
--- a/packages/studio-web/src/app/app.component.spec.ts
+++ b/packages/studio-web/src/app/app.component.spec.ts
@@ -6,7 +6,7 @@ import {
   HttpTestingController,
 } from "@angular/common/http/testing";
 import { TestBed } from "@angular/core/testing";
-import { ReactiveFormsModule } from "@angular/forms";
+import { FormsModule, ReactiveFormsModule } from "@angular/forms";
 import { BrowserAnimationsModule } from "@angular/platform-browser/animations";
 import { RouterTestingModule } from "@angular/router/testing";
 
@@ -23,6 +23,7 @@ describe("AppComponent", () => {
       imports: [
         RouterTestingModule,
         ReactiveFormsModule,
+        FormsModule,
         BrowserAnimationsModule,
         MaterialModule,
         ToastrModule.forRoot(),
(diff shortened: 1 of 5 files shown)


joanise:
As I mentioned in Slack, I need help getting unit testing to pass on this PR. Submitting it anyway, to facilitate reviewing the changes. But we should not merge it until CI is fixed.

joanise:
Converted this PR to a draft.

The audio tweaks pass CI, so I'll submit a separate PR for those, so we can merge them without waiting for the fix on CI for the text tweaks.

dhdaines (approved):
This looks good and uncontroversial, one thing I noticed is that if you navigate away from the page and then back to it, the browser will remember the text in the textbox, but Studio-Web won't.  I'm not sure how to fix that.

joanise:
> one thing I noticed is that if you navigate away from the page and then back to it, the browser will remember the text in the textbox, but Studio-Web won't. I'm not sure how to fix that.

Hum, it's worse than that. If you navigate away and come back, it can be difficult to recover and make the app usable again without reloading. But I don't think that has anything to do with these changes.

PR #100, opened 2023-02-14
Tweak audio recording widget

### LINKED ISSUE #32: When recording fails, we get both an error message and a Yay success message

### LINKED ISSUE #88: Studio-W: while playing audio, there should be a stop button
When you play the audio you just recorded or uploaded in step 1, you can't stop it. The play button should transform into a stop button while it's playing the audio so you can interrupt it.

### LINKED ISSUE #89: Studio-W: warn user to listen to their recorded audio before continuing

joanise:
Several improvements:
 - remove false success message that sometimes happened
 - put a warning to prefer this only for short audio tests
 - while listening to the playback, display a stop button, and reset on end/error/stop

Fixes: #89
Fixes: https://github.com/ReadAlongs/Studio-Web/issues/32
Fixes: #88


packages/studio-web/src/app/upload/upload.component.html
```diff
@@ -72,6 +72,7 @@ <h2 i18n="Title for upload box" class="subtitle">Select an audio file</h2>
                 <div *ngIf="inputMethod.audio === 'mic'" class="mb-0 mb-md-3">
                     <div class="row">
                         <h2 i18n="Title for record box" class="subtitle">Record your voice here</h2>
+                        <p i18n="Warning for record box">Best suited for short tests.</p>
```
roedoejet: I don't think this is the appropriate place for this. Maybe put it as a tooltip or as part of the shepherd tour. I find it confusing to have this as a `p` element above the button.

packages/studio-web/src/app/upload/upload.component.html
```diff
@@ -84,18 +85,23 @@ <h2 i18n="Title for record box" class="subtitle">Record your voice here</h2>
                         Stop
                       </button>
                     </div>
+                  <p *ngIf="!recording && audioControl.value !== null" i18n="playback invitation">Please Play your audio to make sure it is OK, and save it for reuse if it is.</p>
```
roedoejet: Similarly here, I'd rather this was a temporary toast message than markup on the page.

packages/studio-web/src/app/upload/upload.component.html
```diff
@@ -84,18 +85,23 @@ <h2 i18n="Title for record box" class="subtitle">Record your voice here</h2>
...
+                    <button i18n="Audio stop button" class="audioControl" color="warn" (click)="stopPlayback()" *ngIf="!recording && audioControl.value !== null && playing" mat-button aria-label="Stop button">
+                      <mat-icon class="mat-icon-lg">stop_circle</mat-icon>
+                      Stop
+                    </button>
                   </div>
                   <div class="col-12 col-lg-4">
-                    <button i18n="Audio download button" class="audioControl" color="primary" (click)="downloadRecording()" *ngIf="!recording && audioControl.value !== null" mat-button color="primary" aria-label="Audio download button">
-                      <mat-icon class="mat-icon-lg">file_download</mat-icon>
-                      Download
+                    <button i18n="Audio download button" class="audioControl" color="primary" (click)="downloadRecording()" *ngIf="!recording && audioControl.value !== null" mat-button color="primary" aria-label="Audio save button">
```
roedoejet, line 100: Yep, this is better wording!

packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -140,16 +141,22 @@ export class UploadComponent implements OnInit {
   playRecording() {
     if (!this.playing && this.audioControl.value !== null) {
       let player = new window.Audio();
+      this.player = player
       player.src = URL.createObjectURL(this.audioControl.value);
-      player.onended = () => {
-        this.playing = false;
-      };
+      player.onended = () => this.stopPlayback();
```
roedoejet, line 152: Nice! this is the proper way to do it.

roedoejet (changes requested):
I think the messaging around the buttons should be made into toast messages and/or instructions in the shepherd tour.

packages/studio-web/src/app/upload/upload.component.html
joanise: toastr message is a good idea, yes, I agree that would be better.

packages/studio-web/src/app/upload/upload.component.html
dhdaines: I agree, it looks out of place

packages/studio-web/src/app/upload/upload.component.html
dhdaines: Also "Play" maybe shouldn't be capitalized?

dhdaines (approved):
Looks good, Aidan already suggested some appropriate changes

packages/studio-web/src/app/upload/upload.component.html
joanise: Yeah, I guess I'll take it out. I don't like it as a tooltip either:
image
and this is no better:
image

dhdaines (approved):

Issue #49, opened 2022-11-16
assets should not get prepended for a blob


joanise:
When using `<graphic url="data:image/jpeg;base64,<b64 encoded blob>"/>`, even if `use-assets-folder` is true, `assets/` should not get prepended.

deltork:
FIXED in branch patch.page_transition

joanise:
Fixed again in #101

PR #101, opened 2023-02-14
Improved error messages and asset loading

joanise:
Copied over from the page transition branch and adapted to the new state of the code.


joanise:
Fixes #49

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -816,10 +828,12 @@ export class ReadAlongComponent {
...
    *
-   * @param word
-   * @param lang
+   * @param word short name for the text to fetch
+   * @param lang language code
+   * @param path (optional) the path/file/href the error message applies to
+   * @param assetType (optional) type of assert the error message applies to
    */
-  returnTranslation(word: string, lang?: InterfaceLanguage): string {
+  returnTranslation(word: string, lang?: InterfaceLanguage, path?: string, assetType?: string): string {
```
dhdaines: if we are going to support more than 2 languages this way of doing translations will get ugly fast, but for the moment it will be okay...

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1138,6 +1152,8 @@ export class ReadAlongComponent {
     <i class="material-icons-outlined">subtitles</i>
   </button>
 
+  ErrorMessage = (props: {msg: string, data_cy: string}): Element => <p data-cy={props.data_cy} class="alert status-error"><span class="material-icons">error_outline_outlined</span> {props.msg}</p>
```
dhdaines: Can we avoid extra-long lines like this? I find them very hard to read...

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1170,60 +1186,34 @@ export class ReadAlongComponent {
...
-              : "XML"}
-            </span>
-          </p>
-        }
+        {Object.entries(this.assetsStatus).map((asset) => {
+          let assetType = asset[0]
+          let code = asset[1]
+            if ( code === ERROR_PARSING ) {
+              let path = this.getPathFromAssetType(assetType)
+              return <this.ErrorMessage msg={this.returnTranslation("parse-error", this.language, path, assetType)} data_cy={assetType + "-error"}/>
```
dhdaines: again please format this and below to multiple lines...

packages/web-component/test-data/ej-fra/ej-fra-malformed.ras
```diff
@@ -0,0 +1,27 @@
+<?xml version='1.0' encoding='utf-8'?>
+<read-along>
+    <!--this xml is intentionally modified to simulate someone making a mistake with the expected format of this file-->
```
dhdaines: this is, technically, "invalid" rather than "malformed" - there is already an example in ej-fra-invalid.ras I think?

example of ill-formed XML:
```xml
<opening-tag>
</mismatched-closing-tag>
```

Whereas "invalid" means that it doesn't correspond to the DTD or schema, as in this case where a tag is misspelled (I used `<txet>` in my example but `<divs>` is probably a better example, thanks!

dhdaines (approved):
Looks good, the only changes I'd like are formatting-related (can we run prettier on that code?)

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: Yes, I want to run prettier, but it changes about 90% of this file, so not before we've merged in the last feature changes. Anything outstanding will not be rebasable after we run prettier.

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: I do like what prettier does with this line of code:

```
  ErrorMessage = (props: { msg: string; data_cy: string }): Element => (
    <p data-cy={props.data_cy} class="alert status-error">
      <span class="material-icons">error_outline_outlined</span> {props.msg}
    </p>
  );
```

packages/web-component/test-data/ej-fra/ej-fra-malformed.ras
joanise: Your comment is confusing me. I open a `<divs>` that I close with `<div>`, so I think that's what you mean I should call malformed or ill-formed, right?

It might be interesting to add an actual invalid test case too, with valid XML that fails to match the DTD. But I can already tell you that Web-Component silently accepts it, and just skips text from any element that does not follow the expected structure.

dhdaines: ah, oops! never mind, I didn't read until the end of the file.  We do have an invalid test case already, in ReadAlong-Studio, just not in Web-Component... and actually it's good that Web-Component silently skips unknown elements, I think, as that's expected HTML5 behaviour

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: But I will run prettier on just the new lines, that's not going to conflict with anything in the future.

packages/web-component/test-data/ej-fra/ej-fra-malformed.ras
joanise: Good to know silently skipping is what we're supposed to do. I wasn't sure if we should make an issue of it or not, but no, not if it's already doing the right thing.

PR #102, opened 2023-02-15
Option to hide tr at load time

joanise:


dhdaines (approved):
Looks straightforward and works for me

PR #103, opened 2023-02-15
Scroll behaviour option, auto vs smooth

joanise:


dhdaines (approved):
Looks good - I assume the name "auto" comes to us from some underlying component?  It's not very descriptive, but oh well...

joanise:
> I assume the name "auto" comes to us from some underlying component? It's not very descriptive, but oh well...

Indeed, "auto" and "smooth" are the values we pass to whatever actually implements the page scroll animation under the hood.

PR #104, opened 2023-02-15
feat: option to control playback rate range

joanise:
default is "narrow": 85-115
option is "medium": 75-125

When we implement pitch preserving speed control, we can add "wide" with 50-150 or something like that.

The range needs to be symmatric around 100 (e.g., 100 +/- 15) so that 100 is the center of the slider range, otherwise there's no indication to the user on how to set it back to the original speed.


dhdaines (changes requested):
Could we specify the range numerically instead?  I think the values are kind of arbitrary.

Also, would it be possible to have a button that resets it to 100?  I think this should be pretty easy to do...

joanise:
How about `playback-rate-range="15"` to mean 100 +/- 15? "15" would be the default, and the creator could pick the number they wanted?

dhdaines:
> How about `playback-rate-range="15"` to mean 100 +/- 15? "15" would be the default, and the creator could pick the number they wanted?

Yes, that's exactly what I was thinking!

joanise:
Changes pushed, @dhdaines @roedoejet ready for re-review.

I didn't put in any unit test for this, but you can test it manually by editing `test-data/ej-fra/index-playback-rate.html` and changing the `playback-rate-range` value.

If you want a good laugh, set it to 90 and play the readalong at the slowest speed allowed (10% in this case). I sound like a dragon breathing and lightly growling in its sleep.

dhdaines (approved):

PR #111, opened 2023-02-15
fix(fr): create specific configs for serve and use --localize for deployment

roedoejet:
See https://github.com/nrwl/nx/issues/2839

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -26,7 +26,8 @@ jobs:
           npx nx test:once studio-web
       - name: Build
         run: |
-          npx nx build studio-web && npx nx build studio-web --configuration fr
+          npx nx build studio-web --configuration=production
+          npx nx build studio-web --configuration=production --localize=fr --deleteOutputPath=false
       - name: Deploy 🚀
         uses: JamesIves/github-pages-deploy-action@v4.2.3
         with:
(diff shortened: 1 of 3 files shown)


roedoejet:
@dhdaines - this is why the /fr/ endpoint isn't working (I think). In production, we replace the localhost:8000 endpoint with the actual RAS api, but for some reason (argh) nx silently fails to build if you specify more than one configuration (something that is perfectly valid in angular.json). So, as a hack, I've just copy/pasted the production configuration into the fr configuration. This could fix the issue assigned to you on clickup, but there might be a better way of doing it.

dhdaines:
Thanks for the link!  There seem to be various workarounds possible - I'll play around with it a bit to see what our best option is.

dhdaines (approved):
I think it is better to be explicit about what we are doing when we build for deployment - instead of trying to mix configurations, we can just use --localize and explicit --deleteOutputPath=false, which says what we're doing - build a localized version inside the already-built default English one.

dhdaines:
Note that for `serve` you can't pass `--localize` so in this case we just take the Cartesian product of configurations and languages, which is fine for the moment since we only have two languages.

dhdaines:
Whoops, I re-broke API problem!  Will make another PR...

dhdaines:
Oh, never mind, it's just GH pages deployment is slow (because Ruby).  It's fixed now!

PR #113, opened 2023-02-16
Switch from .ras to .readalong as file extension

dhdaines:


joanise (changes requested):
Mostly good, but needs i18n and FR l10n updates too.

joanise:
PS: you edited the text of the messages in `i18n/*` but you also need to update the keys.

joanise:
PPS: I'm fixing it for you, and I'm also going to add CI checks that i18n and l10n are up to date, so we find out faster.

dhdaines:
> PS: you edited the text of the messages in `i18n/*` but you also need to update the keys.

Oh, what a pain!!!  I wonder why Angular i18n couldn't just work like good old gettext which has existed for 100 years and never had this problem.

dhdaines (commented):
No idea how to confirm that the changes are made, I think you have to switch your review to "approve"!  Thanks for the updates!

joanise (approved):
This is ready to merge once CI passes.

PR #114, opened 2023-02-16
eradicate.xml.namespaces

### LINKED ISSUE #112: Downloading other formats doesn't work if you've already downloaded a single file html
There must be something that gets added to the RAS (either image or translations) that the RAS API doesn't like because you can download all file formats if you haven't downloaded a single file html yet. As soon as you download the single file html (which adds translation and image elements to the ras), then the other formats don't work.

dhdaines:
Because we parse RAS as XML, if you insert something that looks like an HTML element into the resulting `Document`, the browser may (this is not documented) helpfully add `xmlns="http://www.w3.org/1999/xhtml"` to its attributes, because it's very important that everybody know that it's XHTML! Even though it isn't!

Explicitly tell the browser to NOT do this, because it is NOT helpful.

Fixes #112

### DIFF
--- a/packages/studio-web/src/app/demo/demo.component.ts
+++ b/packages/studio-web/src/app/demo/demo.component.ts
@@ -45,20 +45,21 @@ export class DemoComponent implements OnInit {
     } else {
       const sentence_nodes = doc.querySelectorAll("s:not(.sentence__translation)")
       // represents all translation nodes that have already been added
-      const translation_node_ids = Array.from(doc.querySelectorAll(".editable__translation")).map((t_node) => t_node.id)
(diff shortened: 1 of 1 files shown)


joanise (approved):
Looks good to me.

roedoejet:
> Because we parse RAS as XML, if you insert something that looks like an HTML element into the resulting `Document`, the browser may (this is not documented) helpfully add `xmlns="http://www.w3.org/1999/xhtml"` to its attributes, because it's very important that everybody know that it's XHTML! Even though it isn't!
> 
> Explicitly tell the browser to NOT do this, because it is NOT helpful.
> 
> Fixes #112

Nice catch!

PR #118, opened 2023-02-16
Translate the component and use the translated version in the french locale

roedoejet:
Does two things:

 - put max size check on text to upload to web-api: cap plain text at 10kb, .ras at 20kb
 - default to french readalong on french site
 - translate tooltips, buttons, and such on french readalong


packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -59,7 +59,9 @@ export class UploadComponent implements OnInit {
   player: any = null;
   progressMode: ProgressBarMode = "indeterminate";
   progressValue = 0;
-
+  maxTxtSize = 10 * 1024; // Max 10 KB audio file size
+  maxRasSize = 20 * 1024; // Max 20 KB audio file size
```
joanise: Not audio, text:
```suggestion
  maxTxtSize = 10 * 1024; // Max 10 KB plain text file size
  maxRasSize = 20 * 1024; // Max 20 KB .readalong XML text file size
```

packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -332,6 +334,10 @@ export class UploadComponent implements OnInit {
         { timeOut: 10000 }
       );
     } else if (type === "text") {
+      if ((file.name.split('.').pop() === 'txt' && file.size > this.maxTxtSize) || (file.name.split('.').pop() === 'readalong' && file.size > this.maxRasSize)) {
```
joanise: I don't think we restrict extensions to just .txt and .readalong.
I would rather do something like this:
```suggestion
      if ((file.name.split('.').pop() !== 'readalong' && file.size > this.maxTxtSize) || (file.size > this.maxRasSize)) {
```
so we don't leave any potential door open.

joanise (changes requested):
Just some small change requests, otherwise looks good.

packages/studio-web/src/app/upload/upload.component.ts
roedoejet: you're right, we currently do accept .xml, but we shouldn't!

joanise: That's not what I mean. I mean `foo.bar` should be accepted and assumed to be a plain text file. Unless we explicitly bar it, but I would find that exceedingly annoying.

joanise:
Just pushed a patch to this PR to tell the user what the max size actually is. There's nothing I hate more than a message that tells me I'm doing something wrong without telling what would have been right!

joanise (approved):
With my latest changes, and David's translations, I think this is good.

PR #119, opened 2023-02-17
feat(tour): add tour for step 2

roedoejet:
This PR adds the tour to the second step of the studio. This code is somewhat painful, since we need to get Shepherd to load demonstration audio/text, interact with the stepper, and also focus in on elements that are part of the web component's shadow dom.

I generated the messages.json but did not add French translations. This PR shouldn't be merged until those are added.


dhdaines (approved):
This is super cool, thanks! I added translations and a warning about the step change which can take some time.

dhdaines:
Sometimes the step transition just hangs forever for me :( I can't figure out why, everything seems to succeed otherwise and there is no error in the logs.

dhdaines:
> Sometimes the step transition just hangs forever for me :( I can't figure out why, everything seems to succeed otherwise and there is no error in the logs.

This doesn't seem to be a problem with this PR specifically, so don't let it block merging.

joanise:
Some text seems to be missing here:
image
the list of formats is introduced but not given. I'd actually just remove that, though, unless we feel the need to explain the formats we support.

joanise (changes requested):
There's an important bug happening when we take the tour for Step 2: if you click Next on the last tour step on Step 1, it reset the text and audio! If you have started adding text or audio, then you lose your stuff and you can't recover it. When you come back to Step 1 after finishing the tour, you have "Hello world!" is the text box and Aidan reading it in the audio. Nothing wrong with that text and recording, except when it erases the user's existing work!

Second bug: if you take the tour twice, the second time it's confused, starting with the Step 2 tour steps, but overlaid on the Step 1 page instead, so it's quite confusing.

Third bug: the keyboard shortcut (right cursor arrow) doesn't work from the last step on Step 1 to the first step on Step 2, and the obvious quick fix (adding `type: "next"` to `step_one_final_step.buttons[1]`) just broke the transition completely for reasons I don't understand.

I have not reviewed the French text yet for all this, but I thought it was worth submitting these bug reports before continuing.

joanise:
The tour for Step 2 would be a good place to warn the users not to use the browser's back button to go back to Step 1. I would add one last step to the tour, with the title "Warning" and the text "Please don't use the browser's back button to go back to Step 1, doing so will erase all your text and audio. Use the Step 1 button instead." 
Fr: "Attention" "Faites attention de ne pas utiliser le bouton de retour en arrière de votre navigateur pour retourner à l'Étape 1, car ça effacerait votre texte et votre audio. Utilisez plutôt le bouton Étape 1."

packages/studio-web/src/app/shepherd.steps.ts
```diff
@@ -145,16 +145,170 @@ export const language_step: any = {
...
+    },
+  ],
+  id: "readalong-export",
+};
+
+export const readalong_go_back_step: any = {
+  title: $localize`Edit your ReadAlong\n`,
+  text: $localize`If you'd like to change your audio or text you can always go back to Step 1 and change things or make a new ReadAlong, but be careful! Going back to the previous step will mean you will lose all of your translations and images.`,
+  attachTo: {
+    element: "div.mat-horizontal-stepper-header-container",
```
joanise, line 326: Already mentioned in Slack, but it's worth noting here in the PR: this selector depends on Angular innards that could change in future versions, so it may not be stable. However, the element we're trying to attach to is a generated one, so I don't see any obvious fix.

joanise (approved):
I'm now ready to approve this, Aidan and I have made the changes I've requested.

But now there are merge conflicts with main... sigh.

joanise:
Merged main into here and fixed the conflicts.

PR #107, opened 2023-02-15
fix: show error toast if ras api is unreachable

roedoejet:
Shows this now when the RAS api is down:

image

It will also stop the loading bar and reset back to a functional state.

PR #120, opened 2023-02-17
Improved and cleaned up error reporting

dhdaines:
Consolidate all the error handling for uploads in a single place, using the error callback from `Observable.subscribe`.  Also handle audio decoding errors (though... we should probably add an mp3 decoder for poor Safari)


packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -38,7 +37,17 @@ import { HttpErrorResponse } from "@angular/common/http";
 })
 export class UploadComponent implements OnInit {
   langs$ = this.rasService.getLangs$().pipe(
-    map((langs: Array<SupportedLanguage> | HttpErrorResponse) => {
+    catchError((err, caught) => {
```
roedoejet: This is already done in ras.service which is where I think it should be implemented.

packages/studio-web/package.json
```diff
@@ -14,7 +14,7 @@
         "test": "ng test",
         "test:once": "ng test --watch=false --browsers ChromeHeadlessCI",
         "extract-i18n": "ng extract-i18n --format json --output-path src/i18n; echo '' >> src/i18n/messages.json",
-        "check-fr-l10n": "bash -c \"diff -w <(cat src/i18n/messages.json | sed 's/:.*//') <(cat src/i18n/messages.fr.json | sed 's/:.*//')\""
+        "check-fr-l10n": "bash -c \"diff -w <(cat src/i18n/messages.json | sed 's/:.*//' | sort) <(cat src/i18n/messages.fr.json | sed 's/:.*//' | sort)\""
```
roedoejet, line 17: This is done in https://github.com/ReadAlongs/Web-Component/pull/118

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -43,29 +56,46 @@ export class DemoComponent implements OnInit {
...
+        )
+      );
       sentence_nodes.forEach((sentence: Element) => {
         // Add a translation
-        if (sentence.id in translations && !translation_node_ids.has(sentence.id)) {
+        if (
+          sentence.id in translations &&
+          !translation_node_ids.has(sentence.id)
+        ) {
           // No namespaces!! NO! NO! NO!
```
roedoejet, line 94: hahah I like this comment.

packages/studio-web/src/app/file.service.ts
```diff
@@ -8,34 +8,24 @@ import { Injectable } from "@angular/core";
...
-  readFileAsData$ = (blob: any) =>
-    Observable.create((obs: any) => {
-      if (!(blob instanceof Blob)) {
-        obs.error(new Error("`blob` must be an instance of File or Blob."));
-        return;
-      }
-
-      const reader = new FileReader();
-
+  }
```
roedoejet, line 20: Nice - these are cleaner now.

packages/studio-web/src/app/ras.service.ts
```diff
@@ -46,41 +45,16 @@ export enum SupportedOutputs {
...
-        err.message,
-        $localize`Hmm, we can't connect to the ReadAlongs API. Please try again later.`,
-        {
-          timeOut: 60000,
-        }
-      ); return of(err)} ),
-      take(1)
-    );
+  convertRasFormat$(body: ReadAlongFormatRequest, output_type: SupportedOutputs|string): Observable<any> {
+      return this.http.post(this.baseURL + "/convert_alignment/" + output_type, body, {responseType: 'blob'});
```
roedoejet, line 51: We should keep the error catching in the service, otherwise the code is not DRY and we have to remember to implement the catching of errors anywhere we request the langs. I see you added this logic in the upload component, but it ought to live in the service imo.

packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -266,30 +280,51 @@ export class UploadComponent implements OnInit {
         ),
       })
         .pipe(
-          tap((joined_audio_and_ras: any) => {if (joined_audio_and_ras.readalong instanceof HttpErrorResponse) {
+          catchError((err, caught) => {
             this.loading = false;
-          } }),
-          takeWhile((joined_audio_and_ras: any) => !(joined_audio_and_ras.readalong instanceof HttpErrorResponse)),
+            throw err;
+          }),
```
roedoejet: Why remove takeWhile? We should unsubscribe from the observable when it throws an error.

roedoejet (changes requested):
I appreciate the prettifying! We're planning on doing this on the whole code base soon too I believe. 

The refactoring of the error catching to being outside of the service is not what we should do I don't think. I'm happy to discuss this on slack, but I would vote to not accept these changes. The error messages seemed to work just fine on the tests that Eric and I did that was PR'ed and reviewed here: https://github.com/ReadAlongs/Web-Component/pull/107. 

The changes to readFile$ and readFileAsData$ are good and make the code more readable as well. I think all the changes in file.service.ts are good to keep. Thanks as always for the work

packages/studio-web/src/app/upload/upload.component.ts
dhdaines: If the observable throws an error, it stops, and reports it to the error callback, no?

packages/studio-web/src/app/upload/upload.component.ts
dhdaines: Perhaps ... this is just ugly in general, I don't like how we are doing all of this logic outside any function, and before the UI is even constructed.

dhdaines: It would be better if we could explicitly subscribe to this observable and *catch* the error

packages/studio-web/src/app/upload/upload.component.ts
dhdaines: It really seemed to me that there was a *lot* of unnecessary code here, and checking the type like this is gross

dhdaines: (I agree that checking the error type in the `error` handler of the subscriber is not so great, but RxJS gives us no other option, and it is at least semantically equivalent to a series of catch blocks for each type of exception)

packages/studio-web/src/app/ras.service.ts
dhdaines: Yes, I disagree with this.  The error catching *and forwarding* in the service makes the downstream code *more* complicated, as it has to have a more complicated type signature.  An error is not a piece of data that should be put into the output stream, it's an exception that should be caught at the appropriate place, which, if we are going to do UI stuff to report it to the user, is in the *component* and not in the service.

packages/studio-web/src/app/file.service.ts
dhdaines: thanks, I found the gist where you got them from originally and noticed that it was actually somewhat wrong :)

packages/studio-web/src/app/demo/demo.component.ts
dhdaines: thanks :) note that it really seems to be some undefined browser behaviour that's causing this problem, I would be interested to see if it happens on Safari or Firefox too

packages/studio-web/package.json
dhdaines: ah, thanks!  I didn't notice it there, I was just annoyed at having to manually move things around

packages/studio-web/src/app/upload/upload.component.ts
dhdaines: Now that I think of it, the real problem here is that this error is *not* recoverable at all!  You have to reload the page to make it work again.  I think this is a special case that can't be treated like the others, we shouldn't even be showing the application given that it absolutely will not work in this case.

dhdaines: I changed this.  `langs` really has no business being an Observable anyway, it isn't a stream of data.  Also, we need to handle more HTTP errors than just failures to connect, and these *are* specific to different situations.  To make the code DRY I consolidated the reporting into a single method.  I think it is all much cleaner now, and we may even be able to recover from `langs` if the API comes back on line (which we couldn't do when it was an Observable referenced directly in the template code)

dhdaines: indeed, we can now actually recover from `langs` failing, although just as `und` for the moment - *but* because we don't want the user to start creating stuff only to find that they are unable to continue and have to reload the page and lose all their data, I think it is best to not show the upload UI in this case, so I changed it to do that.

packages/studio-web/src/app/ras.service.ts
```diff
@@ -46,41 +45,16 @@ export enum SupportedOutputs {
...
-      ); return of(err)} )
-    );
+  assembleReadalong$(body: ReadAlongRequest): Observable<ReadAlong> {
+      return this.http.post<ReadAlong>(this.baseURL + "/assemble", body);
   }
-  getLangs$(): Observable<Array<SupportedLanguage>|HttpErrorResponse> {
-    return this.http.get<Array<SupportedLanguage>>(this.baseURL + "/langs").pipe(
-      catchError((err: HttpErrorResponse) => { this.toastr.error(
-        err.message,
-        $localize`Hmm, we can't connect to the ReadAlongs API. Please try again later.`,
```
dhdaines, line 79: this wasn't very DRY, by the way, duplicating the same error reporting twice just lines apart

dhdaines:
Note that all this improved error handling needs to be tested in CI (I tested it thoroughly myself already...), *but* because we don't have Cypress set up for studio-web, we can't do that yet.  This is actually one of my ClickUp tasks!  But it will have to be in a separate PR.

packages/studio-web/src/app/upload/upload.component.html
```diff
@@ -128,8 +136,8 @@ <h1 i18n="Title for language box" class="title">Select your Language</h1>
                                 <mat-label i18n="Label for language dropdown">Select Language</mat-label>
                                 <mat-select [formControl]="langControl" required>
                                     <mat-option i18n="Name of undetermined language" value="und">Undetermined - (und)</mat-option>
-                                    <mat-option *ngFor="let lang of langs$ | async" [value]="lang['id']">
-                                        {{lang['name']}} - ({{lang['id']}})
+                                    <mat-option *ngFor="let lang of langs" [value]="lang.code">
+                                        {{lang.names['_']}} - ({{lang.code}})
```
roedoejet, line 140: The only problem with this is that `| async` handles proper unsubscribing. I should really refactor all of our subscriptions to unsubscribe when the component is cleared though to repair memory leaks

packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -22,34 +30,14 @@ import {
...
   styleUrls: ["./upload.component.sass"],
 })
 export class UploadComponent implements OnInit {
-  langs$ = this.rasService.getLangs$().pipe(
-    map((langs: Array<SupportedLanguage>|HttpErrorResponse) =>
-    {
-    if (Array.isArray(langs)) {
-      return langs
-      .map((lang) => {
-        return { id: lang.code, name: lang.names["_"] };
```
roedoejet, line 39: are we still sorting? or is this happening in the API?

roedoejet (approved):

packages/studio-web/src/app/upload/upload.component.ts
joanise: It happens in the client, not in the API.

joanise: Unless that was changed recently.

roedoejet: no, it wasn't changed, I just tested it out. this change breaks our sorting-by-name so this should be reverted before accepting this PR

roedoejet: image

joanise: OK, then yeah, we can't remove client-side sorting. I didn't realize this PR did that.

dhdaines: Hi, no, I don't want to *revert* this change because the old code was gross.  Instead I will just, like, sort the languages correctly.  Sorry, I didn't notice that we were doing that here.

dhdaines: Done.

roedoejet: Thanks for fixing @dhdaines! Please avoid assessing code as “gross” in code reviews though.

dhdaines: Ok, no problem! If it's your code I insulted, sorry about that!  I will use other descriptives like "confusing" or "needlessly complex" or "suspicious"

packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -87,6 +75,16 @@ export class UploadComponent implements OnInit {
...
+    this.rasService.getLangs$().subscribe({
+      next: (langs: Array<SupportedLanguage>) => {
+        this.langs = langs.sort((a, b) => {
+          if (a.names['_'] < b.names['_']) return -1;
+          if (a.names['_'] > b.names['_']) return 1;
+          return 0;
+        });
+      },
+      error: (err) => this.reportRasError(err),
+    });
```
joanise, line 87: I have to agree this code is less confusing that what we had before, thanks for cleaning up language sorting. (The sorting part was mine, and I have to admit I was just baffled by how hard it was to write `sorted(my_array, key=lambda x: x.names['_'])` in JS/TS.

I still have a question, though: I thought we were supposed to avoid doing real work like this in constructors, and defer it to `ngOnInit`, so that instantiation of the component for unit testing would not trigger these calls, only actual component initialization would. We were already initializing at construction time before you changes, so my question is orthogonal to this PR, but shouldn't we move this to `ngOnInit()`?

dhdaines: That's a really good question, maybe @roedoejet knows the answer? We need `langs` to be defined before the component gets rendered... so, if `ngOnInit()` happens before rendering, we probably should do it there.

dhdaines: also - it turns out this exists, for string comparison, so perhaps we should use it: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

sorry I called your code gross! it's not you, it's JavaScript :)

dhdaines: My reading of this https://angular.io/guide/lifecycle-hooks is that yes, `ngOnInit` is the right place for it, I will make a PR with that change and the other one I just mentioned.

dhdaines: nope. Angular is aggressive about rendering, so if we load the langs in `ngOnInit` we get a Flash Of Unintented Content with "Server Error" before it finishes loading them.

joanise: Yeah, I saw `localeCompare` when I searched, but I didn't like adding a dependency just for this.

And good to know it has to go in the constructor, thanks!

joanise: Oh, wait, I think it was because `localeCompare` sorts an array of strings, but I needed to dig into the structure for the key, which Python supports trivially with a lambda function, but apparently not JS.

Anyway, yes, JS is confusing. Often!

PR #121, opened 2023-02-17
translate component UI (thanks monorepo!)

### LINKED ISSUE #117: Fix translations in French
Currently there are some missing translations at https://readalong-studio.mothertongues.org/fr/ on Step 2.

image

dhdaines:
There's a bit of overlap with #119 - the way we select the language for the UI is correct here, it does not depend on the URL.

This fixes #117


dhdaines:
Ah, oops, this should actually just be part of #118 - sorry, I will rebase it there

PR #123, opened 2023-02-17
Dev.bundle

roedoejet:
This just adds new build targets for the css and js bundles required for the offline html. We could have done this by including them as assets from the web-component dist and then updating the `nx build web-component` command to include `prepublish` and `bundle` but that broke adding the `--watch` flag and didn't work with `nx serve`. So, instead, we write directly to studio-web assets and include `prepublish` and `bundle` as required steps for publishing studio-web.


dhdaines (approved):
Ah, cool, yes, now that I look at it this is the obvious way to do this!  Thanks!

Ideally we wouldn't have to check the bundles into git under `studio-web/assets` but I imagine it's not obvious to make Nx do the right thing and build them for us...

roedoejet:
yea, my first (naïve) attempt was to get Nx to do it, but that put a little too much faith into Nx. It became very hard to do very quickly, mostly since you have to define all these specific command executors and whether they run parallel etc. I got it to work without `--watch` but that wasn't good enough so this solution was a bit easier.

joanise (commented):
Oh, this is super nice, getting the bundle right from our own assets, thanks for making the change!

PR #125, opened 2023-02-17
Small upload cleanups

dhdaines:
- use `localeCompare` to compare strings more clearly
- be Angularly correct and move `/langs` loading into `ngOnInit`
- delay rendering the upload component until we know if we have langs or not
- report  failure to load the SoundSwallower WASM file to the user

### DIFF
--- a/packages/studio-web/src/app/upload/upload.component.ts
+++ b/packages/studio-web/src/app/upload/upload.component.ts
@@ -1,6 +1,6 @@
 // -*- typescript-indent-level: 2 -*-
 import { ToastrService } from "ngx-toastr";
-import { Observable, forkJoin, of, zip } from "rxjs";
+import { Observable, forkJoin, of, zip, finalize } from "rxjs";
 import {
   map,
   catchError,
@@ -97,17 +97,16 @@ export class UploadComponent implements OnInit {
       console.log(err);
       return;
     }
-    this.rasService.getLangs$().subscribe({
-      next: (langs: Array<SupportedLanguage>) => {
-        this.langs = langs.sort((a, b) =>
-          a.names["_"].localeCompare(b.names["_"])
-        );
-        this.isLoaded = true;
-      },
-      error: (err) => {
-        this.isLoaded = true;
-      },
-    });
+    this.rasService
+      .getLangs$()
+      .pipe(finalize(() => (this.isLoaded = true)))
+      .subscribe({
+        next: (langs: Array<SupportedLanguage>) => {
+          this.langs = langs.sort((a, b) =>
+            a.names["_"].localeCompare(b.names["_"])
+          );
+        },
+      });
   }
 
   reportRasError(err: HttpErrorResponse) {
--- a/packages/studio-web/src/app/upload/upload.component.ts
+++ b/packages/studio-web/src/app/upload/upload.component.ts
@@ -77,24 +77,24 @@ export class UploadComponent implements OnInit {
         $localize`Whoops, something went wrong while recording!`
       );
     });
-    this.rasService.getLangs$().subscribe({
-      next: (langs: Array<SupportedLanguage>) => {
-        this.langs = langs.sort((a, b) => {
-          if (a.names['_'] < b.names['_']) return -1;
-          if (a.names['_'] > b.names['_']) return 1;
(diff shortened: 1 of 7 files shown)


packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -77,24 +85,28 @@ export class UploadComponent implements OnInit {
...
-          if (a.names['_'] > b.names['_']) return 1;
-          return 0;
-        });
-      },
-      error: (err) => this.reportRasError(err),
-    });
   }
 
-  async ngOnInit(): Promise<void> {
+  async ngOnInit() {
```
joanise, line 90: Will the call to `getLangs$()` wait until after ssjs is fully initialized?

Since `getLangs$()` is used as a subscription without waiting, while we await on ssjs, I would have intuitively inverted the two blocks in this function, so they can happen in parallel.

joanise (approved):
Nice refactoring, thanks! Looks good to me, and works on my machine.

I just have one question about the order of the code in `ngOnInit()` (see in line comments) and I tested, the appears to load (trivially) faster when I invert the ssjs and getLangs blocks.

joanise:
The commit graph was getting hard to follow, so I took the liberty to rebase both your PRs onto `origin/main` so these ones create simple graphs when merged.

If you had any outstanding work on them, or if you're going to do changes based on my feedback, please make sure you're working from my rebased version of your branches.

dhdaines:
> > If you had any outstanding work on them, or if you're going to do changes based on my feedback, please make sure you're working from my rebased version of your branches.
> 
> Luckily I don't have any outstanding work! What does this mean, that I just run `git pull` and it should fix everything for me? If it complains, should I delete and re-pull the branch?

Yeah, `git pull` is complaining, and I don't understand what it is saying.  I don't know if I want to set a global option or not:

```
hint: You have divergent branches and need to specify how to reconcile them.
hint: You can do so by running one of the following commands sometime before
hint: your next pull:
hint: 
hint:   git config pull.rebase false  # merge (the default strategy)
hint:   git config pull.rebase true   # rebase
hint:   git config pull.ff only       # fast-forward only
hint: 
hint: You can replace "git config" with "git config --global" to set a default
hint: preference for all repositories. You can also pass --rebase, --no-rebase,
hint: or --ff-only on the command line to override the configured default per
hint: invocation.
fatal: Need to specify how to reconcile divergent branches.
```

packages/studio-web/src/app/upload/upload.component.ts
dhdaines: you're absolutely right!  also, I am going to change it to download the models as well, as I mentioned in #130

dhdaines:
> The commit graph was getting hard to follow, so I took the liberty to rebase both your PRs onto `origin/main` so these ones create simple graphs when merged.
> 
> If you had any outstanding work on them, or if you're going to do changes based on my feedback, please make sure you're working from my rebased version of your branches.

Ok! Got it ... just for the record, your answer to how to do this makes a lot of sense:

> Have you made any additional commits locally after the version I rebased?
> If not, git reset --hard origin/dev.branch-name will do the trick.
> If you have, git rebase origin/dev.branch-name.

packages/studio-web/src/app/upload/upload.component.ts
dhdaines: This is a bit more complicated to do correctly, actually, though I will do it - the issue is that we want to correctly handle the error where SoundSwallower cannot be loaded in the same way we handle a failure to contact the API.  The right way to do this is, I think to make an `Observable` (or a `Subject`, whatever) for the initialization and do it in "parallel" with the `/langs` loading, then catch the errors for both of them.

dhdaines: Automatically testing a SoundSwallower load failure is something I'm not sure how to do either - currently I temporarily replace `soundswallower.web.wasm` with an empty file and rebuild everything.

dhdaines: this is done, perhaps not in the prettiest or most Angularly correct way, but in a way that avoids a less-than-typesafe Subject or a global error handler

PR #127, opened 2023-02-18
Large microphone cleanup

dhdaines:
The microphone service code had a lot of unused and redundant stuff in it.  This makes it much simpler and also simplifies the upload code that uses it, improving error handling slightly.


packages/studio-web/src/app/microphone.service.ts
```diff
@@ -1,132 +1,64 @@
...
-    return MicrophoneService.guc();
-  }
-
-  startRecording() {
-    if (this._recorderState === RecorderState.RECORDING) {
-      this.recorderError.emit(ErrorCase.ALREADY_RECORDING);
-    }
-    if (this._recorderState === RecorderState.PAUSED) {
+  async startRecording() {
+    if (this.recorder !== null && this.recorder.state == "paused") {
```
joanise, line 12: `=== "paused"`?

joanise (approved):
Wow, that's nice. So much simpler with async that tracking the states. Thanks for that change!

packages/studio-web/src/app/microphone.service.ts
dhdaines: Hi! The states are defined here https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/state (which is why it didn't make much sense to replicate them in the service ...)

dhdaines:
> Wow, that's nice. So much simpler with async that tracking the states. Thanks for that change!

Thanks, my Mac is supposed to arrive today, so I will make sure everything works on Safari before merging

packages/studio-web/src/app/microphone.service.ts
joanise: Sorry, I finally realized my question was unclear. I was asking about `==` vs `===` and I'm thinking we want three equal signs here.

PR #139, opened 2023-02-20
Dev.memory

roedoejet:
use the pattern of defining an `unsubscribe$` subject on each component that emits and completes when the component is destroyed. all subcriptions in components should then use the `takeUntil` operator to complete when unsubsribe emits.


packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -85,7 +86,10 @@ export class UploadComponent implements OnInit {
       langs: this.rasService.getLangs$(),
       _: this.ssjsService.waitForInit$(),
     })
-      .pipe(finalize(() => (this.isLoaded = true)))
+      .pipe(
+        finalize(() => (this.isLoaded = true)),
```
dhdaines, line 90: I'm a bit confused by this - if `finalize()` was already working here (it was), doesn't that mean that the Observable completed successfully and the subscription ended?

Do you have a reference for this pattern and the reason it's necessary?  I'm not totally convinced...

packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -294,7 +303,8 @@ export class UploadComponent implements OnInit {
             // We can't give the arguments types because RxJS is broken somehow,
             // see https://stackoverflow.com/questions/66615681/rxjs-switchmap-mergemap-resulting-in-obserableunknown
             this.ssjsService.align$(audio, ras as ReadAlong)
-          )
+          ),
+          takeUntil(this.unsubscribe$)
```
dhdaines, line 307: Likewise - we are definitely calling the `complete` callback on the subscription here, and in fact there is an explicit unsubscribe callback that we return when creating the `Observable` which I can also confirm is getting called and destroying the decoder.

dhdaines (changes requested):
Not necessarily changes but I would like a reference to the reason we need to do this...

dhdaines:
Actually just a follow-up comment.  It seems to me that the reason why a subscription would leak is if it can't be garbage collected.  This would be the case if we stored it in an attribute or a global variable, as in the examples [here](https://levelup.gitconnected.com/preventing-angular-subscription-memory-leaks-65f18eb9cb8e).  But in most (all?) of these cases we are not actually doing that, so I would assume that it gets garbage collected when it is no longer accessible from any code, which would be the case once it is complete.

packages/studio-web/src/app/upload/upload.component.ts
roedoejet: I think this is because of forkJoin which completes after the first emission. So, you're right that it's not necessary here. as a rule of thumb, I just apply this pattern to each subscription because it doesn't hurt and because applying it selectively would require analyzing each subscription as to whether there's a `take(1)` buried in the service somewhere or knowing that `forkJoin` completes after a single emission etc. I can try and dig up some documentation on the pattern

dhdaines: I see what you mean, it doesn't hurt or complicate the code unnecessarily, and covers the case where, for whatever reason, the component gets destroyed when a subscription is still outstanding...

roedoejet: Yea, it looks like the only place this might happen is when we call `this.rasService.convertRasFormat$` - everything else either has `forkJoin` or `take(1)` applied to it seemingly. But, ya, as a rule of thumb I don't think it's a bad pattern to explicitly unsubscribe, and it gives us an easy way of doing so in the future if we introduce more observables that don't complete on their own.

dhdaines (approved):

PR #141, opened 2023-02-20
fix: add warning if nothing was recorded

### LINKED ISSUE #136: Recording fails silently
Sometimes recordings look like they've happened, but there's only silence and there is no warning given to the user. This seems to happen on my laptop when my headphones aren't plugged in - I suspect this has something to do with the browser choosing the wrong input or something, but I think we should at least have a warning if we don't record any input.

roedoejet:
fixes https://github.com/ReadAlongs/Web-Component/issues/136

Needs i18n for French


packages/studio-web/src/app/microphone.service.ts
```diff
@@ -33,7 +33,11 @@ export class MicrophoneService {
     return new Promise((resolve, reject) => {
       this.recorderEnded.subscribe(
         (blob) => {
-          resolve(blob);
+          if (blob.size < 2000) {
```
dhdaines, line 39: we could probably make this even smaller - it would be good to have a comment explaining why it works

packages/studio-web/src/i18n/messages.json
```diff
@@ -109,8 +109,9 @@
     "4183225119057268962": "Could not start recording!",
     "2596823344081631983": "Audio was successfully recorded. Please listen to your recording to make sure it's OK, and save it for reuse if so.",
     "1317075918959775059": "Yay!",
-    "1983793909601149790": "Please try again, or select a pre-recorded file.",
```
dhdaines, line 112: OK, I can translate :)

dhdaines (changes requested):
I'll add some translations quickly!

packages/studio-web/src/app/microphone.service.ts
roedoejet: 1500? I struggled to get the size less than 3000 (ie by clicking record/stop record as fast as I could) if the microphone was working

dhdaines (approved):

roedoejet:
@dhdaines the CI failed unfortunately

PR #142, opened 2023-02-20
Studio web improvements from Delaney's feedback

joanise:
This PR implements those changes that easy to do.

 - change text of record button while recording
 - make text of re-record button explicitly say delete and re-record
 - make next step button more contrastive
 - fix the padding around the save text button


roedoejet (approved):
Look good to me thanks @joanise !

PR #145, opened 2023-02-20
fix: hide output format on mobile devices

### LINKED ISSUE #129: file format output dropdown is not visible on mobile
You can't see the dropdown on mobile because of the spacing. I would recommend we just don't display the dropdown on mobile - why would anybody want to download an ELAN file on their phone? Better to de-clutter the UI on mobile and hide it on small devices.

roedoejet:
fixes https://github.com/ReadAlongs/Web-Component/issues/129

### DIFF
--- a/packages/studio-web/src/app/app.component.ts
+++ b/packages/studio-web/src/app/app.component.ts
@@ -64,6 +64,11 @@ export class AppComponent implements OnDestroy, OnInit {
       $localize`Warning`,
       { timeOut: 10000 }
     );
+    // User Browser's default messaging to warn the user when they're about to leave the page
+    window.addEventListener("beforeunload", (e) => {
+      if (this.formIsDirty()) (e || window.event).returnValue = true;
+      return true;
+    });
   }
 
   ngOnDestroy(): void {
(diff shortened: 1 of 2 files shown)


packages/studio-web/src/app/demo/demo.component.html
```diff
@@ -12,7 +12,7 @@
         </h2>
       </div>
       <div class="col-3 download__buttons">
-        <mat-form-field appearance="fill">
+        <mat-form-field class="d-none d-lg-block" appearance="fill">
```
joanise, line 15: That's a good idea, it makes no sense to allow all the output formats on a mobile device.

Question of alignment: I would like the download icon to be at the far right when the download types are not there. I tried to edit `demo.component.sass` to do it, but I failed to move that darn thing. Not critical, but it would be nice to have the download__buttons div float to the right if you can figure out how to do it right.

packages/studio-web/src/app/app.component.ts
```diff
@@ -64,6 +64,11 @@ export class AppComponent implements OnDestroy, OnInit {
       $localize`Warning`,
       { timeOut: 10000 }
     );
+    // User Browser's default messaging to warn the user when they're about to leave the page
```
joanise, line 67: Sweet!

joanise (approved):
Good stuff. My requested change about the local of the icon is esthetic, would be nice but not required.

PR #146, opened 2023-02-20
refactor: move loading of model to happen on application load

roedoejet:
then check for model loaded when aligning.

I'm not totally sure about this implementation, but the basic idea is there, that we shouldn't wait to load the page for soundswallower to be initialized. It will take the user a little while to create the readalong and by that time, their connection may have loaded soundswallower and related assets anyways. On slow connections, we give an error to ask the user to try again.

TODO: Requires French translations please, CI won't pass until that


packages/studio-web/src/app/demo/demo.component.sass
```diff
@@ -3,3 +3,4 @@
     button.download
         margin-top: 5px
         margin-left: 15px
+    justify-content: flex-end
```
joanise, line 6: Oy vey! IU tried with flex and padding and align and margin. Very CSS. There's one trivial thing that's needed, but good luck figuring out which one!

joanise (approved):
Looks good. I'm not sure how I could exercise this new error, but the code looks fine.

dhdaines:
Hi, I didn't get a chance to review this because I got stuck in some other work this afternoon.  I will make a new PR on top of it, because there are some changes I'd like to see - in particular if loading SoundSwallower fails, the user doesn't find out until they try to align, which isn't good, because there's no possibility to recover from that error.

roedoejet:
No worries! Sounds good. I'll review your PR when it comes. Cheers

PR #147, opened 2023-02-21
Dev.remove page title editing

roedoejet:
I expect this change to be something we can talk about at the meeting tomorrow. I appreciate the work that went into this, and it was very well tested, but I think at this point as we develop more and more features, it introduces an unnecessary amount of complexity into the interface. I vote that we remove this functionality.


roedoejet:
note, this is based on https://github.com/ReadAlongs/Web-Component/pull/145 so the only relevant commit here is [a13a66c](https://github.com/ReadAlongs/Web-Component/pull/147/commits/a13a66c202ce4fa4b5fa78a51108e362da5800cd)

joanise:
Something is broken with this code, when I try to align it systematically tells me it can't load the audio in step 2.

packages/studio-web/src/app/demo/demo.component.html
```diff
@@ -35,20 +35,7 @@
...
-          class="subtitle"
-          [(ngModel)]="slots.pageTitle"
-          [ngStyle]="{ 'width.ch': slots.pageTitle.length, 'min-width.ch': 25 }"
-          style="border: none"
-          (input)="onPageTitleChange($event)"
-          placeholder="Enter your page title here"
-        />
-      </div>
-    </div>
+    <div class="row"></div>
```
joanise, line 38: Why is this row still here if it's empty?

joanise:
> Something is broken with this code, when I try to align it systematically tells me it can't load the audio in step 2.

Never mind, I can't reproduce the problem anymore. It works fine now.

joanise:
I'm not quite sure why we want to delete this, so let's talk about it at the meeting. 

One related thing that's been bugging me, the shepherd tour, when it says you can edit the title and subtitle here, it only highlights the title. If we could make it highlight both the title and the subtitle together, that would be nicer. But of course, the trivial way didn't work for me, I tried to add a div around them and find it by id, but adding that div kills the functionality altogether. So this is only a would-be-nice.

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -159,7 +159,7 @@ export class DemoComponent implements OnDestroy, OnInit {
       <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0">
-        <title>ReadAlongs</title>
+        <title>${this.slots.title}</title>
```
joanise, line 162: Wow, that was nicely trivial!

joanise (approved):
Beauteous!

PR #148, opened 2023-02-21
Support Safari 15 and newer

dhdaines:
This supports recent versions of Safari, simply by polyfilling MediaRecorder to record in mp3 format.

There are some other small changes, which should have allowed us to supporter older versions, but we can't, because Angular refuses.

If you try to compile to ES2015, it will fail to define `$localize` when doing localized builds (because in fact this function isn't used after AOT compilation when tagged templates are fully supported, but when they aren't, it is still necessary, but Angular does not know or believe this).

If you somehow get past that problem, there is an error deep in its guts that I can't possibly understand, involving an `UnsupportedError` in `executeListenerWithErrorHandling` whatever that is.


joanise (commented):
Frankly, I don't understand most of the changes in this PR, so if you want a real technical review you'll need someone else's, but it works well on my machine, and it's nice to have mp3. So I'm good with this.

joanise (approved):

roedoejet (approved):
Looks good to me

PR #150, opened 2023-02-21
Download audio with correct extension

dhdaines:
Just a quick fix to dev.backward-browsers!


roedoejet (approved):
looks good

joanise:
Good catch on this, I failed to notice when I tested on my PC with Chrome that I was getting a WebM file named .mp3, not an actual MP3 file. Thanks for fixing it.

PR #151, opened 2023-02-22
feat: analytics & meta tags for social media and search engines

roedoejet:
Meta tags are required for proper indexing in search engines and also for generating properly formatted sharable links on social media. This is all the ones I could think of/find, but feel free to add more if needed. I localized them and import the homepage from the `package.json`.

Note: i18n is broken again and this needs French.


dhdaines (approved):

roedoejet:
Thanks @dhdaines ! I'm going to update the demo image before merging.

joanise (approved):
Nice! I'm so glad you know how to do this stuff, it's not something I would have thought of!

joanise:
Good idea to use the homepage from package.json. I wonder if it's possible to fetch it from the deployment info, so that it would pick up whatever the base URL is on the server it happens to run on ― that would avoid requiring reconfiguration if we moved it or if someone else deployed their own copy, but it's still just one place that needs editing, so it's certainly good as is.

joanise:
Using Plausible is a good idea, as it does not collect any private information, only general information.
Their data policy makes it clear what they will collect for us: https://plausible.io/data-policy

We should add to our Privacy policy that we collect such usage statistics and link to the Plausible data policy in our own privacy policy.

joanise:
Just to make sure no one else does it at the same time, I'm doing the translation now. Looks good otherwise.

packages/studio-web/src/app/app.component.ts
```diff
@@ -258,7 +258,8 @@ export class AppComponent implements OnDestroy, OnInit {
 
   openPrivacyDialog(): void {
     this.dialog.open(PrivacyDialog, {
-      width: "250px",
+      width: "50vw",
+      maxHeight: "90vh",
```
joanise, line 262: Just wondering, what's a vw and what's a vh?

joanise (approved):
Nice, looks good.

packages/studio-web/src/app/app.component.ts
roedoejet: view width and view height. 1vX is 1% of the viewport width or viewport height. so it's instead of pixels (which on some screens don't nicely correspond with viewport size...looking at youuuu iPhones 😄 ). I should be more consistent in my use of it maybe, but I usually use it just in these cases for large containers and things like that

PR #152, opened 2023-02-22
Improved preloading and fatal error handling

dhdaines:
- catch truly fatal errors earlier and report them on the main page instead of toastrs
- make sure to delete the decoder we use in preloading to avoid memory leak (oops)


joanise (changes requested):
This is nice. I was able to see both the Aligner error (by corrupting my `node_modules/soundswallower/model/en-us/`) and the Server error (by killing my serve-web-api), or either separately, or neither, and it's both clear.

Small change I would do: 
 - "Please contact the developers" should say "Please contact the developers and send them a complete copy of the following log".
 - with serve-web-api not running, the log displays "Http failure response for http://localhost:8000/api/v1/langs: 0 Unknown Error" but the console shows me "net::ERR_CONNECTION_REFUSED" which is a lot more informative. Might the code displaying the error on screen be able to fetch that real error message instead of "0 Unknown Error"?

dhdaines:
> * "Please contact the developers" should say "Please contact the developers and send them a complete copy of the following log".

Sure, I will make that change now!

> * with serve-web-api not running, the log displays "Http failure response for http://localhost:8000/api/v1/langs: 0 Unknown Error" but the console shows me "net::ERR_CONNECTION_REFUSED" which is a lot more informative. Might the code displaying the error on screen be able to fetch that real error message instead of "0 Unknown Error"?

Unfortunately I can't figure out where the `ERR_CONNECTION_REFUSED` message is actually coming from ... it's not present in the error object that Angular gives us, and the message is only printed somewhat later, seemingly after the Observable we get from the HTTP client fails and terminates...

dhdaines:
Yeah, it seems that connection refused errors are reported *deep inside the browser* somewhere that ordinary JavaScript cannot reach without the Ring of Power (no, just kidding, it really just is not accessible to JS, for security reasons, it seems): https://stackoverflow.com/questions/28556398/how-to-catch-neterr-connection-refused

joanise (approved):
Thanks for the message edit.

Too bad we can't capture the real error for langs, but oh well, we can live with it. We'll likely know anyway, and find out quickly when we test on our side.

roedoejet:
Nice! these errors are handled nicely @dhdaines . I think I would prefer to see a redirect or error page though. Currently, the user can still see part of the app (the stepper, and tour button for example) and they can take the tour (which is broken). I'm thinking something more along the lines of: https://www.spotify.com/us/404/ or https://bitly.com/a/404notfound (except for our errors, just 404s)

roedoejet (changes requested):
Errors are caught properly, but as commented, I think we should have an error page for this explicitly.

dhdaines:
> Nice! these errors are handled nicely @dhdaines . I think I would prefer to see a redirect or error page though. Currently, the user can still see part of the app (the stepper, and tour button for example) and they can take the tour (which is broken). I'm thinking something more along the lines of: https://www.spotify.com/us/404/ or https://bitly.com/a/404notfound (except for our errors, just 404s)

Yes, that makes a lot more sense, it feels a bit hacky to be conditioning everything with `*ngIf="fatalError"`.  I'm not totally certain how to do this from an already-loaded web page though...

dhdaines:
Oh, that's right, minimally one can just assign to window.location :) but presumably we want that to be an Angular route of some sort.

roedoejet:
> Oh, that's right, minimally one can just assign to window.location :) but presumably we want that to be an Angular route of some sort.

Yep, exactly.

roedoejet:
@dhdaines want me to set up the routing and you tackle the missing page or?

roedoejet:
I added a PR onto this branch that has stubs for the fatal error and 404 pages https://github.com/ReadAlongs/Web-Component/pull/158

dhdaines:
Thanks! I didn't have time to get to it yet today, will take a look at your PR now.

roedoejet:
> Thanks! I didn't have time to get to it yet today, will take a look at your PR now.

Are you able to join our zoom call?

PR #154, opened 2023-02-22
Disable recording button while we are trying to start to record

joanise:
Another UX tweak: when I first click on Record after leading the app, it can take a few seconds to start, during which it's not clear anything is happening. Solution: disable the button so it's grayed out while we're trying to launch the recorder.

Also, "Hourra!" just sounds too childish to me, it bugs me every time I see that message. I switched it to "Bravo!" which is a little better.


roedoejet (approved):
Looks great! Nice improvement @joanise

dhdaines (approved):
looks good!

PR #164, opened 2023-02-23
feat: rename Save button to Save a copy for better clarity

### LINKED ISSUE #160: Change the "save" button
Some users might think that they *have* to save in order to proceed.

image

joanise:
This PR is fine for the most part, except I don't like how the button is displayed on narrow screens:

image
image

Fixes #160


roedoejet:
Hm, I agree but I don't see an easy way around it.

roedoejet (approved):
This is an improvement, even with the minor style bug in French on mobile.

PR #167, opened 2023-02-24
chore: add meta tags for seo

roedoejet:
dynamically generated meta tags aren't crawled by social media robots


joanise (commented):
I can already see this morning that this is working well, because a link to our site on Slack shows the example image and the description.

PR #169, opened 2023-02-24
fix: clarify what the language selection is actually for

### LINKED ISSUE #159: Change language in studio
"Select the language of your text and audio" instead of "Select your Language"

joanise:
A test user imagined it might trigger translation. While reading the page carefully would have fixed that, we want the first impression to be more accurate instead.

Fixes #159


roedoejet (approved):
Thanks @joanise !

Issue #124, opened 2023-02-17
"contact us" link currently goes nowhere


dhdaines:
Ideally we should pull it from package.json

joanise:
I'm not sure I want a real e-mail address in here. How about a link to a project page? Which page would be our logical project page? https://readalongs.github.io/ICLDC-Docs/ would be a find choice during ICLDC, I guess.

dhdaines:
Yeah, that makes sense, we can always change it later!

joanise:
We now use `readalong.studio@gmail.com` as the contact us e-mail address.

PR #170, opened 2023-02-24
Replace contact us by project page link

joanise:
Related to #124, but not quite fixing it yet.

I'd like the URL I'm linking to to come from package.json or project.json. I don't know how to do that, but I hope it's trivial for one of you.


roedoejet:
the whole package.json is available from `environment` so you just `import { environment } from 'environments/environment'` (the folder is right next to i18n). In production the environment.ts file gets swapped out for environment.prod.ts so don't worry about pulling environment from there or anything

joanise:
Fixed. Ready for re-review.

joanise:
Oh wait, I put my variable definition in `environment.ts`, is that OK, or should I literally put it in `package.json`?

joanise:
I think I mean it now. :)

dhdaines (changes requested):
Hi! It needs the contact info updated on the error page too, otherwise looks good - perhaps with a link specifically to the issues page.

joanise:
> Hi! It needs the contact info updated on the error page too, otherwise looks good - perhaps with a link specifically to the issues page.

Right, how about making "contact the developers" a link to that documentation page? I assume we'll eventually put contact information on there...

joanise:
All right, version one million, I now use the new e-mail address Marc created and the contact links are mailto: links.

The documentationURL is still in the package.json, but not shown anywhere. Post ICLDC, we can fix that and have a real general documentation page.

dhdaines (approved):

PR #171, opened 2023-02-24
Tweak the list of accepted audio formats

### LINKED ISSUE #161: Change language around accepted file formats
remove "webm" from text, add "m4a" to accepted file types.

joanise:
…M4A and WEBM

.webm is an obscure format, most user will not know what it is. We have to support it since the Save button generates it on Windows, but it's enough to support it. Having it in the `accept` list means it'll be avaiable to select so it should not be (too) confusing.

I'm also adding .m4a to the `accept` list (without mentioning it in the text) because on Windows, that's the default format you get when you record, and it's the format of my sample in the Studio.git repo that found its way everywhere, so we really should allow selecting it!

Fixes #161

### DIFF
--- a/packages/studio-web/src/app/upload/upload.component.html
+++ b/packages/studio-web/src/app/upload/upload.component.html
@@ -134,7 +134,7 @@
             i18n="Instructions for uploading audio"
             for="updateAudio"
             class="form-label"
-            >Use pre-recorded audio from an mp3, wav, or webm file.</label
+            >Use pre-recorded audio from an MP3 or WAV file.</label
           >
           <input
             (change)="onFileSelected('audio', $event)"
@@ -142,7 +142,7 @@
             name="audio"
             type="file"
             id="updateAudio"
-            accept=".mp3,.wav,.webm"
+            accept=".mp3,.wav,.webm,.m4a"
           />
         </div>
         <div *ngIf="inputMethod.audio === 'mic'" class="mb-0 mb-md-3">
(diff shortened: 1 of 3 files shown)


packages/studio-web/src/app/upload/upload.component.html
```diff
@@ -134,15 +134,15 @@ <h2 i18n="Title for upload box" class="subtitle">
             i18n="Instructions for uploading audio"
             for="updateAudio"
             class="form-label"
-            >Use pre-recorded audio from an mp3, wav, or webm file.</label
+            >Use pre-recorded audio from an MP3 or WAV file.</label
```
roedoejet, line 137: I'm not a fan of caps here I don't think

packages/studio-web/src/app/upload/upload.component.html
```diff
@@ -134,15 +134,15 @@ <h2 i18n="Title for upload box" class="subtitle">
...
           >
           <input
             (change)="onFileSelected('audio', $event)"
             class="form-control"
             name="audio"
             type="file"
             id="updateAudio"
-            accept=".mp3,.wav,.webm"
+            accept=".mp3,.wav,.webm,.m4a"
```
roedoejet, line 145: 👍

roedoejet (approved):
would prefer not all caps, but if you feel strongly about it, I'm fine.

packages/studio-web/src/app/upload/upload.component.html
joanise: Well, if you search online, the file extensions are lowercase, but the formats are generally upper case, which is why I changed it.

PR #173, opened 2023-02-24
fix(webc): fix height issue by removing max height

### LINKED ISSUE #168: Demo-component can get two scroll bars instead of just one
With #166, 
the bug reported in #153 is much improved, but now there are potentially two scroll bars when the vertical height of the screen is small:
image

While it's possible to see everything, it's very confusing the figure out where you have to scroll to get the right behaviour. Better would be to force the demo.component to always occupy all the space it needs in its parent component, and have only the parent component have a scroll bar.

On a computer, you see the two scroll bars like in my screen shot, but on a cell phone you have to guess where you're supposed to drag to get scrolling where you want it to go.

The same thing happens with the downloaded single-file HTML. Making my viewport even smaller, I was able to have three competing scroll bars
image

roedoejet:
fixes https://github.com/ReadAlongs/Web-Component/issues/168


joanise (approved):
works beautifully now

PR #181, opened 2023-02-28
Detect and report g2p warnings

dhdaines:
Gives a nice and helpful warning box with all the fallback/g2p failures in the log.  Note that:

- There could be a *lot* of warnings
- Not sure how to put line breaks between them (it doesn't seem possible, unfortunately)

### DIFF
--- a/packages/studio-web/src/app/ras.service.ts
+++ b/packages/studio-web/src/app/ras.service.ts
@@ -14,6 +14,7 @@ export interface ReadAlong {
   parsed: string | null;
   tokenized: string | null;
   g2ped: string | null;
+  log: string | null;
 }
 
 export interface ReadAlongRequest {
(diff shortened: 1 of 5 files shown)


joanise (approved):
This is super nice, thanks! I would made keep the message there longer than 15 seconds, though, if there are a lot of warnings, that's not longe enough to read them. I think 30 seconds would not be too long for this.

dhdaines:
Sounds good, I'll put it up to 30 seconds and also go grep for other possible g2p errors it might report

joanise:
Oops, I just pushed the 30s, I didn't read your reply soon enough.

dhdaines:
No problem.  I think matching all lines in the log with "g2p" in them should get us everything relevant, based on my reading of the ReadAlongs code, so I think we can merge this now!

joanise:
Good idea to get all the g2p errors. I see I should still generate an additional one when the final g2p for a word is empty, though. That's still the cause behind the alignment failure. But that's not a problem for this PR, which is ready to merge as far as I'm concerned.

PR #183, opened 2023-02-28
feat: add versioning to single file bundle

roedoejet:
if people have a single file html that doesn't work, we print out the bundle and the timestamp it was created (automatically generated during bundle process). This will help us determine bugs more easily in single file html formats. we should add something similar to https://github.com/ReadAlongs/Studio


joanise (approved):
Looks good. The only issue is that  `pinea` occurs in the bundle, or `joanis` if I run it on my machine, but that's not a new problem.

PR #195, opened 2023-03-03
Fix release workflow

joanise:
- submit a PR for the updated files, since we cannot push to release or main
 - Remove Node 12 dependencies and other warnings

Fully tested in a separate repo.


roedoejet (approved):
Awesome @joanise - thanks for testing in another repo.

PR #197, opened 2023-03-08
feat(fonts): use bc sans by default

roedoejet:
load from shadow dom as per https://github.com/ionic-team/stencil/issues/2072

also removes requirement for parent applications to load google fonts separately


packages/web-component/src/index.html
```diff
@@ -10,11 +10,6 @@
 
     <script type="module" src="build/web-component.esm.js"></script>
     <script nomodule src="build/web-component.js"></script>
-
-    <link
-      href="https://fonts.googleapis.com/css?family=Lato|Material+Icons|Material+Icons+Outlined"
-      rel="stylesheet"
-    />
```
joanise, line 17: This change does not need to be reflected in the HTML that the Studio CLI produces, does it?
I think it doesn't, because if I understand this patch correctly, we bundle all our fonts, but I want to make sure.

joanise: I take it back, you've removed these lines from the sample files in `test-data/ej-fra`, so `MINIMAL_INDEX_HTML_TEMPLATE` in `Studio/readalongs/text/util.py` will need the same change when we make it depend on the web-component version that includes this PR.

Although I assume it's harmless to leave it in there when we don't use it, so it won't be urgent. But we should make an issue in the Studio repo so we remember to do this when it's time.

joanise (commented):
Testing on my machine the Material-Icons fonts don't work anymore for this patch on the web component, for the play/rewind/stop etc buttons at the bottom of the component.

packages/web-component/src/index.html
roedoejet: Yea, you're right - it's harmless to leave it in, but it's unnecessary. Let's remember to take it out. With this patch, the fonts shouldn't need to be declared in the parent (but it doesn't matter if they are, it'll ultimately just make for a redundant call to the API)

joanise (approved):
Works now, all good!

PR #207, opened 2023-06-14
feat: add aligner fallback settings for difficult alignments

roedoejet:
Here's a PR for implementing the same loosening of beam parameters as found in Studio cli. This is missing French and Spanish translations, and should be tested. I tested it locally with the Ende language files that were not working previously, but this ties in to the larger problems with testing for Studio-Web, and would also require test data that is open source (unlike the Ende example).


dhdaines:
Oh, cool, thank you!  I just haven't had the time to look into this.  The implementation seems clear and simple, I'll review it a bit more closely soon.

dhdaines (approved):
Looks good! I guess this is what the `retry` operator is for :)

joanise:
Tested on my machine, finally. Nicely done, it works well. I would just change the setting back to strict each time the user clicks on "next step".

joanise:
And it looks like you already did that, so good to merge.

joanise (approved):
Nicely done, except

PR #208, opened 2023-06-20
clean up package.json and project.json for studio-web

joanise:
We had broken scripts left in packages/studio-web/package.json that only worked before this was a monorepo. I've deleted those lines, migrated others to project.json, removed duplicates, and added some documentation to help the new developer, and the returning developer who doesn't remember how things are setup three months later (i.e., me!).

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ If your Studio sandbox is in a sibling directory to this sandbox, and you Python
 
 Alternatively run together as:
 
-    nx run-many --targets=serve-test-data,serve-web-api,serve --projects=web-component,studio-web --parallel 5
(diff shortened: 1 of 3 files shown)


dhdaines (approved):
Thank you! This definitely helps - is there a command (npm, nx, etc) that will print out some information on the different targets?

joanise:
> is there a command (npm, nx, etc) that will print out some information on the different targets?

Oh, that's a good question. I am not aware of any, unfortunately.

joanise:
https://nx.dev/packages/nx/documents/show#project claims that `nx show project studio-web` ought to do it, but it doesn't work for me.

roedoejet (approved):
Looks good - thanks @joanise !

joanise:
Good idea to update the run-many, thank @roedoejet

PR #211, opened 2023-06-26
fix: show the toast for digit errors again

joanise:
solution details: throw unhandled errors back out for downstream handling

### DIFF
--- a/packages/studio-web/src/app/upload/upload.component.ts
+++ b/packages/studio-web/src/app/upload/upload.component.ts
@@ -421,7 +421,7 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
               }
               return throwError(() => err);
             } else {
-              return of(err);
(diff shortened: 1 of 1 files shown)


packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -421,7 +421,7 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
               }
               return throwError(() => err);
             } else {
-              return of(err);
+              return throwError(() => err);
```
roedoejet, line 424: Hm. but doesn't this cause `retry` to trigger? I'll test this.

roedoejet: yea, on my machine this triggers the error message three times since any error thrown will cause `retry` to trigger:

image

joanise:
Replaced by #212

PR #212, opened 2023-06-26
Dev.fix

roedoejet:


joanise (approved):
Wow, I love the preview, thanks for enabling that!

I don't understand the code change itself, to be honest, but I've tested both in your preview and locally on my machine, and it solves the problem in both cases, so approved.

Yes, we can close #211 and keep this solution instead.

PR #214, opened 2023-06-28
Close the microphone stream when done recording (fixes #213)

### LINKED ISSUE #213: Studio-Web does not stop using the microphone when it stops recording
Testing on an iPad, my kid noticed the app is still recording when you stop recording. More precisely, the app is still using the microphone, even though we're not recording anymore. On Chrome, you see the same thing:
image

and that can be taken as suspicious activity. We should figure out how to stop the microphone once the user clicks on stop recording.

Before you start recording, it shows this:
image
and that's what it should go back to once we click stop.

dhdaines:
Previously the browser tab looked as if we were still recording.  We don't want the user to think we are spying on them!

### DIFF
--- a/packages/studio-web/src/app/microphone.service.ts
+++ b/packages/studio-web/src/app/microphone.service.ts
@@ -7,14 +7,15 @@ export class MicrophoneService {
   private chunks: Array<Blob> = [];
   private recorder: MediaRecorder | null = null;
   private recorderEnded = new EventEmitter<Blob>();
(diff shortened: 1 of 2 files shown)


joanise (approved):
Nice work, and thanks for doing it so quickly!

Plus, I love the preview, so we can all test it without having to deploy locally. Thanks again for enabling that, Aidan!

joanise:
Fixes #213

PR #217, opened 2023-07-04
Limit height of image to page height

dhdaines:
This seems to effectively fix the issues with displaying images on iPads, and also enforces what I think is a reasonable height on uploaded images, namely, 55vh, which is in theory also the same maximum height as a page (it seems to make them a bit smaller because nobody can comprehend where the size of an element actually comes from in CSS)


joanise:
Nice, thanks for tackling this! 

On Studio-Web, things are better, though not perfect yet.
On the downloaded Web Component, though, it's not fixed. I wonder if your patch is failing to get triggered in that case?

Studio-Web screen shots:

This one is great, with the whole tall image visible:
image

This one is even taller, 1100 pixels tall, and still gets cropped a bit at the bottom (you should completely see my two hand-drawn corners at the bottom):
image

Web-Component screen shots:

On the downloaded Web-C, the original problem persists:
image
image

packages/web-component/src/scss/modules/_pages.scss
```diff
@@ -56,7 +55,7 @@ $max-height: 98vh;
       position: relative;
       @extend %themed-bg;
       img {
-        height: 100%;
+        max-height: $page-size + 0vh;
```
joanise, line 58: I hate CSS! A huge amount of work, for a one-line change in the end!

joanise:
Same results on the iPad: in Studio-Web, I can now see my tall image. But in the downloaded html, viewed in Edge on the iPad, it's still blank.

joanise:
To make sure that wasn't the problem, I checked, and `assets/bundle.js` did get updated in the PR preview, so that's not it.

dhdaines:
> To make sure that wasn't the problem, I checked, and `assets/bundle.js` did get updated in the PR preview, so that's not it.

Hm. The only way to know for sure is to inspect the image in the browser to see if the new CSS is getting picked up.  I was having  a *lot* of problems getting nx to actually rebuild things, some bogus caching was getting in the way.

joanise:
I'm surprised you were having trouble. I now just run this command:

    npx nx run-many --targets=serve-test-data,serve-web-api,serve,serve-fr,serve-es --projects=web-component,studio-web --parallel 6

and everything starts reliably for me.

joanise:
Right, yeah, in Studio-Web:
image

In downloaded Web-C:
image

joanise:
OK, I can confirm it was because the bundle did not get updated: it contains the old code: 
    
    ... .page__col__image img{height:100%;max-width:95%;object-fit:contain}...

joanise:
And it's because we need to run

    npx nx bundle web-component

and commit the results!

Doh!

joanise:
Although the workflow that creates the PR Preview does run that command, and so do our publish and release workflows, so I don't quite understand why we have to do it by hand.

joanise:
Argh! I updated the bundle. It's now fixed locally. But not on the PR preview. Help! @roedoejet

joanise:
Got it! It's because we're fetching the bundle from the live version, not the PR preview:
image

It's only a bug in the PR preview, then, which is fetching assets from `/assets/...` rather than `/pr-previous/pr-217/assets/...`

joanise (approved):

PR #219, opened 2023-07-05
fix: use relative URLs for all assets

### LINKED ISSUE #218: Assets are loaded from `/assets` even when deploying to a subfolder
https://readalong-studio.mothertongues.org/es/ fetches https://readalong-studio.mothertongues.org/assets/bundle.js instead of https://readalong-studio.mothertongues.org/es/assets/bundle.js

Problem noticed reviewing #217, where the PR preview was fetching https://readalong-studio.mothertongues.org/assets/bundle.js instead of the updated https://readalong-studio.mothertongues.org/pr-preview/pr-217/assets/bundle.js

The models and .js files and other resources are fetched relative to the folder where things are deployed, but all the files in `/assets` are fetched from the root of the web server instead of from the correct relative path.

joanise:
Fixes #218

### DIFF
--- a/packages/studio-web/src/app/b64.service.ts
+++ b/packages/studio-web/src/app/b64.service.ts
@@ -11,8 +11,8 @@ import { Segment } from "soundswallower";
   providedIn: "root",
 })
 export class B64Service {
-  JS_BUNDLE_URL = "/assets/bundle.js";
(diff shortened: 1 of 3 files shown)


roedoejet (approved):
Thanks @joanise - this seems like the right thing to do

PR #220, opened 2023-07-06
Three small changes: play/pause tooltip, mat icon typo, aria recommendation

joanise:


joanise:
Note: I'm not sure my Spanish translation is OK, with "Déle play/pause", but I was adapting from the existing text. I'm hoping for a native speaker review before merging.

dhdaines (approved):
I will defer to a native speaker for the Spanish translation (which I find a bit shocking! where's the equivalent of the OQLF?) but aside from that this seems quite sensible.

PR #227, opened 2023-10-23
Dev.release1.1.1

joanise:
Note: the diff is huge because 1.1.0 was not done via a merge to release. The changes from 1.1.0 to 1.1.1 are actually rather trivial.

Creating this new release hoping that doing so will fix the Studio tests that fail, in `Studio/test/test_package_urls.py` which currently fail because unpkg resolve Web-Component@^1.0.0 to 1.1.0 but unpkg doesn't seem to have the two assets we use.


joanise:
Not done right. Will try again.

PR #238, opened 2024-02-26
ci: bump production node version to 18 since 16 is eol

joanise:


joanise:
Note: I'm using node 18 here even though node 20 is available, because Angular 15 declares compatibility with node up to 18 only.

PR #240, opened 2024-03-14
feat(web-component): added auto pause feature

deltork:
A feature to pause the audio at the end of each page was added. For now, it is activated via an attribute


packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -877,6 +884,15 @@ export class ReadAlongComponent {
         } else {
           this.images[i] = null;
         }
+        //load the ids for the end of the page
+        if ("paragraphs" in page) {
+          const paragraphs = (page as Page).paragraphs;
+          const sentence =
+            paragraphs[paragraphs.length - 1].querySelector("s:last-of-type");
+          const word = sentence.querySelector("w:last-of-type");
+          this.endOfPageTags[word.id] = parseFloat(word.getAttribute("dur"));
```
joanise: This is really weird, why is the duration of the last word becoming the ID for the end of page tag?

joanise (commented):
Just a comment for now as I try to understand the code. I'll test it and see if it works well for me, and maybe Marc can test it too.

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -938,6 +954,20 @@ export class ReadAlongComponent {
...
+            if (
+              this.autoPauseAtEndOfPage &&
+              el_tag in this.endOfPageTags &&
+              this.finalTaggedWord !== el_tag
+            ) {
+              //wait for the audio of the last word
+              if (!this.isAutoPaused)
+                setTimeout(
+                  () => this.pause(),
+                  this.endOfPageTags[el_tag] * 800
```
joanise: What does this 800 constant value mean?

packages/web-component/src/components/read-along-component/read-along.tsx
deltork: I am creating a list of the last word spoken on each page, using the ID as the key and the duration of the word to help calculate when to pause the audio.

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: Just tested, and it seems to work well, except this constant is funky: we're setting the timeout at 800ms per second of duration or the last word in the page. If we need to stop before the end of the last word, I'd rather see us doing something like `this.endOfPageTags[el_tag] * 1000 - 50` and document why we want to stop 50ms (or whatever number we choose) before the end of this last word.

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: I understand now, you're using this duration to determine when to pause exactly.

joanise (changes requested):
The code change seems good, modulo my comments above about the exact delay, but this breaks unit testing. `npx nx test:once web-component` fails in two of the six specs we run. You can see the logs at https://github.com/ReadAlongs/Web-Component/actions/runs/8281122034/job/22672362698 which I reproduced locally as well.

packages/web-component/src/components/read-along-component/read-along.tsx
deltork: It is the best "guess estimate" to help convert the duration of the word into a timeout value. Ideally, I would convert the duration directly to milliseconds; however, this bit of code activates during the silence before the beginning of the word in the audio. Maybe Aidan knows the exact point in the silence at which the event is emitted. Then, we can use that in a formula for the timeout value.

deltork:
> The code change seems good, modulo my comments above about the exact delay, but this breaks unit testing. `npx nx test:once web-component` fails in two of the six specs we run. You can see the logs at https://github.com/ReadAlongs/Web-Component/actions/runs/8281122034/job/22672362698 which I reproduced locally as well.

fixed.

joanise:
I got it to bug with a four page readalong, I've emailed you the files I used.

PR #241, opened 2024-03-14
feat: add web bundle as download option

roedoejet:
Here's a first stab at this. A few things:

- [x] The [web preview](https://readalong-studio.mothertongues.org/pr-preview/pr-241/) should work if you want to try it out (select "web bundle" from the download format options)
- [ ] ~I don't love how much logic is in the `demo.component.ts`~ I can live with this for now, but I think we're due for some refactoring
- [x] Maybe we should come up with a better basename than just `output` since if people are combining multiple readalongs they'll have to rename everything. timestamps?
- [x] The `assets` folder stuff needs to be improved. We have the `use-assets-folder` property, but the value is just a boolean. I think we should instead have that be a path that could get prepended to all images, because right now, the images are hardcoded in the `.readalong` file as being in an `assets` folder relative to the location of the ReadAlong, which is sure to break in many deployments. Update: see https://github.com/ReadAlongs/Web-Component/pull/243
- [x] looks like we need to run `prettier` on everything. I didn't do that here yet, to make sure the content changes are easily parsable.
- [x] Translations for french and spanish still have to be added


joanise:
More comments, now that I have tested this feature.

File names:
 - `readalong.zip` would be better than `output.zip`, more self-documenting.
 - Ditto the files in `assets/` in the archive, `assets/readalong.{mp3,readalong}` would be better than `assets/output.*`.
 - `output.txt` is confusing, since it's really the input text. `readalong.txt` would be OK, or maybe `input.txt`.

The `README.txt` file should be outdented against the left margin. My guess is you have it indented probably because of the file where you're defining its contents, but I don't like the result, it ought to start on the first line (remove the blank line) and have no indents. It should also be wrapped at <=79 characters. I know we don't do that anymore in code, but here when I double clicked on the file the display was awful at 80 chars wide.
image
I would also be OK with no manual wrapping at all, letting the auto-wrapping do its thing, but pre-wrapped at 107 yields poor results if the window is even slightly narrower. Between not pre-wrapped at all, and pre-wrapped at 79, I can't decide what I prefer: not pre-wrapped adjusts to whatever window size the user has, 79 is old machine friendly, though maybe not mobile friendly... You pick!
Key is, I'd like the default display to be easy to read on all OSes, machine ages, and user defaults.

joanise:
I tested unzipping the file on my machine and running `python -m http.server`, and that works nicely.

One thing, though, this is a zip bomb, with no inner folder. Would it make sense to put all the contents in a `readalong/` folder inside the zip file? Or is it just because I'm a developer that I'm allergic to tar bombs?

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -35,10 +44,30 @@ export class DemoComponent implements OnDestroy, OnInit {
   selectedOutputFormat: SupportedOutputs | string = "html";
   language: "eng" | "fra" | "spa" = "eng";
   unsubscribe$ = new Subject<void>();
+  xmlSerializer = new XMLSerializer();
+  readmeFile = new Blob(
+    [
+      `
+    Web Development Guide
```
joanise: Web Deployment Guide?

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -27,6 +35,7 @@ export class DemoComponent implements OnDestroy, OnInit {
   };
   outputFormats = [
     { value: "html", display: $localize`Offline HTML` },
+    { value: "zip", display: $localize`Web Bundle` },
```
joanise, line 39: Zipped Web Bundle?
That would give me "Fichiers Web zippés" and "Archivos web comprimidos".
I find just "Web Bundle" insufficiently clear.

joanise: I did some research on "zippés" vs "comprimés" for the French translation, and Google tells me the former is much more common than the latter, and that matches my intuition, but Spanish doesn't seem to have that, it would have to be "Archivos web comprimidos zip" if we wanted to be precise, but I think that's unnecessary, adding more verbosity with no additional clarity, whereas the French "zippés" is shorter and more common, thus a no brainer.

joanise:
By the way, we already applied prettier:

```
$ npx prettier -c packages
Checking formatting...
All matched files use Prettier code style!
```

the generated `index.html` file is not prettier'd, though, that's true.

dhdaines:
> The `README.txt` file should be outdented against the left margin. My guess is you have it indented probably because of the file where you're defining its contents, but I don't like the result, it ought to start on the first line (remove the blank line) and have no indents. It should also be wrapped at <=79 characters. I know we don't do that anymore in code, but here when I double clicked on the file the display was awful at 80 chars wide.

`black` wraps at "90-ish" so we do this in code, more or less.  I think it is reasonable to wrap at 80 (if you use Emacs it is as simple as typing `Esc-q`) as this makes the text easier to read in general (there's a reason why LaTeX makes wide margins).  If you don't wrap lines then not everbody's terminal/browser/whatever is going to actually word wrap correctly in any case.

packages/studio-web/src/app/demo/demo.component.ts
dhdaines: compressés: https://vitrinelinguistique.oqlf.gouv.qc.ca/fiche-gdt/fiche/8399261/compresser

dhdaines: mais même selon l'OQLF "zippé" est acceptable dans certains contextes ;-)

dhdaines:
> One thing, though, this is a zip bomb, with no inner folder. Would it make sense to put all the contents in a `readalong/` folder inside the zip file? Or is it just because I'm a developer that I'm allergic to tar bombs?

Yes, I think this is a good idea - MacOS or Windows are going to unpack it in a new folder anyway, but this way nobody gets any nasty surprises.

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -189,6 +235,74 @@ export class DemoComponent implements OnDestroy, OnInit {
...
+            <body>
+                <!-- Here is how you declare the Web Component. Supported languages: en, fr -->
+                <read-along href="assets/${basename}.readalong" audio="assets/${basename}.mp3" theme="light" language="en">
+                    <span slot='read-along-header'>${this.slots.title}</span>
+                    <span slot='read-along-subheader'>${this.slots.subtitle}</span>
+                </read-along>
+            </body>
+
+            <!-- The last step needed is to import the package -->
+            <script type="module" src='https://unpkg.com/@readalongs/web-component@^1.0.0/dist/web-component/web-component.esm.js'></script>
```
dhdaines: Should we instead make sure this is the same version as the web application?  I think we can get it from the environment during build

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -189,6 +235,74 @@ export class DemoComponent implements OnDestroy, OnInit {
...
+        </html>
+        `;
+      const indexHtmlFile = new Blob([sampleHtml], { type: "text/html" });
+      zipFile.file("index.html", indexHtmlFile);
+      // - add plain text readme
+      zipFile.file("README.txt", this.readmeFile);
+      // - write zip
+      zipFile
+        .generateAsync({ type: "blob" })
+        .then((content) => saveAs(content, `${basename}.zip`));
```
dhdaines: Could this promise be rejected?  We should handle possible errors here.

packages/studio-web/src/app/soundswallower.service.spec.ts
```diff
@@ -583,12 +588,19 @@ AFwAWQBbAGEAagBvAHAAcABkAGIAaABhAGcAagBwAGkAVwBfAA==`;
...
   let service: SoundswallowerService;
-  let audioService: AudioService;
-
+  let fileService: FileService;
+  let httpClientSpy: jasmine.SpyObj<HttpClient>;
+  let httpClient: HttpClient;
+  let httpTestingController: HttpTestingController;
   beforeEach(() => {
-    TestBed.configureTestingModule({});
+    httpClientSpy = jasmine.createSpyObj("HttpClient", ["get"]);
```
dhdaines, line 596: Jest testing seems to work with nx and angular now... when I do the angular 17 migration I will eradicate Jasmine for good

packages/studio-web/src/app/upload.service.ts
```diff
@@ -0,0 +1,11 @@
+import { Injectable } from "@angular/core";
+import { BehaviorSubject } from "rxjs";
+
+@Injectable({
+  providedIn: "root",
+})
+export class UploadService {
+  $currentAudio = new BehaviorSubject<Blob | File | null>(null);
+  $currentText = new BehaviorSubject<Blob | string | null>(null);
```
dhdaines, line 9: What does the convention say to do with the `$` anyway?  Before or after the name?  Does it depend if it's a `Subject` or an `Observable`?

packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -83,7 +84,7 @@ export class UploadComponent implements OnDestroy, OnInit {
     audio: "mic",
     text: "edit",
   };
-  textInput: string = "";
+  textInput = new BehaviorSubject<string>("");
```
dhdaines: Since this is an `rxjs` thingy it should have a `$` ... somewhere ... in the name to be consistent.

packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -92,11 +93,23 @@ export class UploadComponent implements OnDestroy, OnInit {
...
     private rasService: RasService,
     private fileService: FileService,
-    private audioService: AudioService,
     private ssjsService: SoundswallowerService,
     private microphoneService: MicrophoneService,
+    private uploadService: UploadService,
     private dialog: MatDialog
-  ) {}
+  ) {
+    this.audioControl.valueChanges
```
dhdaines: Likewise with all of these rxjs creatures

dhdaines (changes requested):
Thank you!  Looks good, mostly style comments.

packages/studio-web/src/app/upload.service.ts
roedoejet: I thought a $ prefix was subject and a $ suffix was observable. But it looks like Angular only explicitly documents the use of $ as a suffix for observables.

dhdaines: Hmm - now that I think of it, `Subject`s are also `Observable`s, so that makes sense.

roedoejet: true, but I kind of like having $ prefix since it lets you know that you can call `next` on it etc.. but I don't feel super strongly about it. I just changed it to `$textInput` but I can change it if you prefer

dhdaines: Makes sense!  I agree it's useful to have a $ on it so we know what it is...

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -189,6 +237,88 @@ export class DemoComponent implements OnDestroy, OnInit {
       element.click();
       document.body.removeChild(element);
       this.registerDownloadEvent();
+    } else if (this.selectedOutputFormat === "zip") {
+      let zipFile = new JSZip();
+      // Create inner folder
+      const innerFolder = zipFile.folder("readalong");
+      const assetsFolder = innerFolder?.folder("assets");
+      const timestamp = Date.now();
```
joanise: This gives us a timestamp in seconds since epoch, e.g., `readalong-1710864009240`.
Can we do yyyymmddhhmmss instead? It's just one digit longer but it's human readable.

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -35,10 +44,31 @@ export class DemoComponent implements OnDestroy, OnInit {
...
+
+This bundle has everything you need to host your ReadAlong on your own server.
+
+Your audio, (optional) image, and alignment (.readalong) assets are stored in the assets folder.
+
+The plain text used to create your ReadAlong is also stored here along with an example index.html file.
+
+Your index.html file demonstrates the snippet and imports needed to host the ReadAlong on your server.
+
+Please host all assets on your server, include the font and package imports defined in the index.html in your website's imports, and include the corresponding <readalong> snippet everywhere you would like your ReadAlong to be displayed.
```
joanise: That should say `<read-along>`, not `<readalong>`. It look me a while to notice this, but it's been nagging me, something felt wrong but I didn't quite know what... :)

joanise (changes requested):

packages/studio-web/src/app/demo/demo.component.ts
joanise: I'm not going to be picky, but my first choice is "zippé" because we're creating an actual .zip file.

Does anyone have an opinion on the English, though, whether we should add "Zipped" in front?

roedoejet: I think "zipped" is redundant - Bundle implies that it's bundled/zipped which is sufficient

joanise (approved):
Looks good to me now.

dhdaines (approved):
looks good! sorry for the delay in approving!

deltork:
can the readme contain section about deployment on wordpress, with the apporpiate snippet :
   ` [wp_read_along_web_app_loader  version='^1.2.0']
    <read-along...></readalong>
    [/wp_read_along_web_app_loader]`

joanise:
> can the readme contain section about deployment on wordpress, with the apporpiate snippet :
> ` [wp_read_along_web_app_loader version='^1.2.0'] <read-along...></readalong> [/wp_read_along_web_app_loader]`

Absolutely, just add that in your PR.

deltork:
> > can the readme contain section about deployment on wordpress, with the apporpiate snippet :
> > ` [wp_read_along_web_app_loader version='^1.2.0'] <read-along...></readalong> [/wp_read_along_web_app_loader]`
> 
> Absolutely, just add that in your PR.

The readme is only in this branch. I will have to add the instructions after this branch is merged

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -189,6 +237,91 @@ export class DemoComponent implements OnDestroy, OnInit {
...
+    } else if (this.selectedOutputFormat === "zip") {
+      let zipFile = new JSZip();
+      // Create inner folder
+      const innerFolder = zipFile.folder("readalong");
+      const assetsFolder = innerFolder?.folder("assets");
+      const timestamp = new Date()
+        .toISOString()
+        .replace(/[^0-9]/g, "")
+        .slice(0, -3);
+      const basename = `readalong-${timestamp}`;
```
deltork: When people provide a title, adding the title to the filename will help them find it after download, especially if they do multiple books.
`` const basename = (this.slots.title?this.slots.title.replace(/\s/g,"_").toLowerCase():'readalong')+`-${timestamp}`; ``

deltork (approved):

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -160,6 +209,13 @@ export class DemoComponent implements OnDestroy, OnInit {
     if (this.selectedOutputFormat == "html") {
       await this.updateImages(ras);
       await this.updateTranslations(ras);
+      const timestamp = new Date()
+        .toISOString()
+        .replace(/[^0-9]/g, "")
+        .slice(0, -3);
+      const basename =
+        (this.slots.title ? slugify(this.slots.title, 15) : "readalong") +
```
joanise, line 217: This is great, except it's too anglocentric: if the title is ᐃᓄᒃᑎᑐᑦ, the slug is empty and you get `-20240402212031.zip`, and the .txt file inside is similarly limited. Similarly "Éric" becomes "ric", but the Inuktitut example is a lot more dramatic. Can we slugify the text-unidecode output, maybe?

roedoejet: thanks for this @joanise - I've updated it to use unidecode

PR #243, opened 2024-03-19
fix(assets)!: get rid of useAssetsFolder boolean and replace with imageAssetsFolder string

roedoejet:
BREAKING CHANGE: useAssetsFolder is no longer valid.

The equivalent is imageAssetsFolder='assets/' which is the default.

It used to be that all readalong and audio files were forced to be in the same `assets` folder along with the images. Then we removed that requirement, and toggled the image assets as a boolean. But, we should really go one step further and define an imageAssetsFolder property on the readalong so that the `.readalong` file can just define image names and then prefix them with the `imageAssetsFolder` otherwise we'll be asking users to edit the `.readalong` file when they deploy their readalongs.

IMPORTANT NOTE: we need to publish v 1.2.0 to npm before merging this to main, otherwise the bundles that are downloaded from studio will incorrectly point to v1.1.0. Before merging this, please ping me to publish the new version on npm!

Also, I don't think the deploy preview will really work here for the bundle/single file html etc, since v1.2.0 won't be published to unpkg yet


packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -96,11 +96,13 @@ export class ReadAlongComponent {
   @Prop() cssUrl?: string;
 
   /**
-   * Toggle the use of assets folder for resolving urls. Defaults to on
-   * to maintain backwards compatibility
+   * Toggle the use of an assets folder. All image paths
+   * will be prefixed with this. Defaults to 'assets'.
```
joanise, line 100: Not quite accurate, it actually defaults to `'assets/'`, with the slash.

In fact, since you do `this.imageAssetsFolder + path;` in `urlTransform()`, the slash is mandatory. That should be documented if it stays that way, but in fact I would much rather add it automatically if it's not there rather than require it.

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -208,11 +210,12 @@ export class ReadAlongComponent {
...
   private urlTransform(path: string): string {
     if (
-      this.useAssetsFolder &&
+      this.imageAssetsFolder &&
       looksLikeRelativePath(path) &&
       !path.startsWith("blob")
-    )
-      return "assets/" + path;
+    ) {
+      return this.imageAssetsFolder + path;
```
joanise, line 220: As mentioned above, that's a guaranteed gotcha. The intuitive thing to do in the read-along element is `image-assets-folder="images"` and that's going to generate e-mail enquiries for support.
I'd rather implement the logic 
 - if folder is empty or undefined return path
 - else return folder + "/" + path
This else logic is safe even if the user adds the slash, because doubling the slash won't break the URL.

joanise (changes requested):
Good idea to bump the versions for this breaking change.
I'm good with these changes, modulo not requiring the slash.

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -96,11 +96,13 @@ export class ReadAlongComponent {
...
-   * Toggle the use of assets folder for resolving urls. Defaults to on
-   * to maintain backwards compatibility
+   * Toggle the use of an assets folder. All image paths
+   * will be prefixed with this. Defaults to 'assets'.
+   * .readalong files should just contain base filenames
+   * not the full paths to the images.
    */
 
-  @Prop() useAssetsFolder: boolean = true;
+  @Prop() imageAssetsFolder: string = "assets/";
```
deltork, line 105: make it mutable `@Prop({mutable: true}) imageAssetsFolder: string = "assets/";`
so that you can append the trailing slash if missing

packages/web-component/src/components/read-along-component/read-along.tsx
deltork: In `async componentWillLoad()` you can check for the trailing slash and add if needed
`
if (!this.imageAssetsFolder.endsWith("/")){
      this.imageAssetsFolder+="/"
    }
`

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -208,11 +210,12 @@ export class ReadAlongComponent {
    */
   private urlTransform(path: string): string {
     if (
-      this.useAssetsFolder &&
+      this.imageAssetsFolder &&
```
deltork, line 213: Not necessary since `this.imageAssetsFolder` will always exist

deltork (changes requested):
I do not think this will be a breaking change since the asset folder default is the same.

I love the idea of passing a folder to the component! no more hard coding URL in .readalong file for those with custom deployments

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: Is that still going to be true if the html file specifies `images-assets-folder=""`?

joanise:
I just tested this PR, and the Web Bundle I got had 

    <script type="module" src='https://unpkg.com/@readalongs/web-component@^1.1.1/dist/web-component/web-component.esm.js'></script>

where I believe it will need `^1.2.0`. 

We have something a little tricky here: when Studio-Web and Studio-CLI start producing snippets with `^1.2.0`, that will be broken until we publish that on npmjs.

Should we pause auto-updating github-pages from all pushes to main here until we publish web-component 1.2.0 to npmjs?

We're entering an unstable dev stage, that may be the safest way to go. We could change `publish.yml` to run on push to `main_disabled` in this PR and revert that change when everything is ready.

joanise:
> We're entering an unstable dev stage, that may be the safest way to go. We could change `publish.yml` to run on push to `main_disabled` in this PR and revert that change when everything is ready.

Or maybe a better solution is to merge to a `dev` branch for now, and put all breaking changes there until we're ready instead of in `main`.

packages/web-component/src/components/read-along-component/read-along.tsx
deltork: If we append `/` to the `this.imageAssetsFolder` bit, then `images-assets-folder=""` would effectively become `images-assets-folder="/"`. A user can also do `images-assets-folder=".'` equivalent to `images-assets-folder="./"`.

joanise: But that would be a bug, no? If it's `""`, we should not add `/` in front.

joanise (approved):
Looks great!

packages/web-component/src/components/read-along-component/read-along.tsx
deltork: I agree; however, it is an input error. So we need to define how to handle it. If a user provides  `image-assets-folder=""`  it be treated as `"./"`  or `"/"`. I vote `"./"`

PR #244, opened 2024-03-20
Auto pause end of page

deltork:
When the flag `auto-pause-at-end-of-page` is set to `true,` the web component. The audio automatically pauses at the end of each page after the last word is spoken. Clicking the play button resumes the audio on the next page.


packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -938,6 +961,31 @@ export class ReadAlongComponent {
...
+              el_tag in this.endOfPageTags &&
+              this.finalTaggedWord !== el_tag
+            ) {
+              //console.log(this.audio_howl_sprites.sound.seek() ,el_tag, this.endOfPageTags[el_tag], this.play_id);
+              //this is a work around to ignore id's not close to current sound time
+              //
+              const timeDiff = Math.round(
+                (this.audio_howl_sprites.sound.seek() -
+                  this.endOfPageTags[el_tag][0]) *
+                  100 //in microsecond
```
joanise: I haven't tested yet, but this constant of 100 is surprising for what seems to be converting milliseconds to microseconds. Why not 1000?

joanise:
Tested. The last file pair I sent you by e-mail still doesn't work properly, sorry.

packages/web-component/src/components/read-along-component/read-along.tsx
deltork: I will fix the comment to centi-seconds and not microseconds. I emailed you a verbose I used to guestimate the constant. As I was annotating it, 100 centiseconds seemed too large for an error or margin. I think it should be half.

joanise:
Tested again, and with the latest changes it seems to work correctly. This is probably ready to turn back into a non-draft PR.

Small request: please rebase it onto `origin/main`, it looks like you cherry-picked a couple commits from main into your branch, but we don't want to have those twice in the history, it's better to rebase a PR's branch onto main instead and force push that.

PR #247, opened 2024-03-26
Migrate all the things to Angular 17 and Nx 18

### LINKED ISSUE #239: Angular 15 is EOL in May 2024, so bump to Angular 17.
I did a first attempt just using the automatic tool to do 15->16, and it was not straightforward, there are interactions with other dependencies. So this will not be an easy task, but we need to tackle it at some point because we'll be stuck unable to handle CVEs soon enough if we don't.

dhdaines:
This is complete insofar as the tests appear to be passing for me (may need to fix them in CI).

Note that all of the dev dependencies have been consolidated at the top-level package.json, and numerous redundant dependencies have been removed.

Why on earth Angular comes as 50 different packages all of which have to be updated separately I will never know.  What I have learned from this is that:

1. First you migrate nx, then you migrate @angular/core, then you migrate @angular/material.
2. nx is so proud of the fact that it runs migrations separately from updating packages but this is an utterly useless feature since you can only run one migration at a time!  Do not try to accumulate multiple migrations in the same migrations.json!  Do not try to migrate multiple packages on the same command line!  THIS DOES NOT WORK!
3. So instead of one command you need to run like six of them.  But it mostly works except for the npm dependency nightmare.

Also since MacOS likes to run things on port 5000 I moved the test data to a more random-looking port number.

Also, we don't really use @angular/cli for anything except running the 'test:once' script, and this is because we use the very harmful karma/jasmine test framework, so we should switch to jest as it is better, but that's a different PR.

Also, Nx 17 introduced a very annoying breaking change which makes releasing packages from a monorepo needlessly difficult: https://github.com/nrwl/nx/issues/19989 so I have added some hacks which I hope will work, but seriously... why is monorepo good again?  Is it like monorail?


.prettierignore
```diff
@@ -12,3 +12,5 @@ node_modules/
 .angular/
 www/
 packages/ngx-web-component/**/generated
+
+/.nx/cache
```
joanise, line 16: Is this meant to have a leading slash?

joanise:
Sweet: 
```
$ npm audit
found 0 vulnerabilities
```

joanise (approved):
Very nice, thank you for taking this on!

.prettierignore
dhdaines: Yes, apparently it means to only match `.nx/cache` in the top-level directory ... I just became aware of this syntax (also in .gitignore) recently.

(it was added automatically by some Nx migration)

packages/ngx-web-component/package.json
```diff
@@ -1,30 +1,15 @@
...
-    "@angular/compiler": "^15.2.4",
-    "@angular/core": "^15.2.4",
-    "@angular/forms": "^15.2.4",
-    "@angular/platform-browser": "^15.2.4",
-    "@angular/platform-browser-dynamic": "^15.2.4",
-    "@angular/router": "^15.2.4",
-    "rxjs": "~7.5.0",
-    "tslib": "^2.3.0",
-    "zone.js": "^0.11.8"
+    "@angular/core": ">=13",
```
roedoejet, line 7: I assume you confirmed that it works with 13 and 14 as well?

package.json
```diff
@@ -2,48 +2,66 @@
...
+    "@angular/compiler": "^17",
+    "@angular/core": "^17",
+    "@angular/forms": "^17",
+    "@angular/material": "^17",
+    "@angular/platform-browser": "^17",
+    "@angular/router": "^17",
+    "bootstrap": "^5.2.3",
+    "file-saver": "^2.0.5",
+    "image-conversion": "^2.1.1",
+    "ngx-toastr": "^16.1.0",
```
roedoejet, line 58: I've seen this with nx, where all the dependencies end up in the main `package.json` but I don't understand it. How does it know to not bundle `ngx-toastr` with the web-component for example?

.github/workflows/release.yml
```diff
@@ -32,6 +32,10 @@ jobs:
       - run: cd ./packages/web-component && npm version patch
       - run: cd ./packages/ngx-web-component && npm version patch
       - run: npx nx build web-component && npx nx bundle web-component && npx nx build ngx-web-component
+        # We have to do this now because nx is broken (see https://github.com/nrwl/nx/issues/19989)
+      - run: |
```
roedoejet, line 36: gah - that's unfortunate.

roedoejet (commented):
Thanks @dhdaines ! I just asked a few clarifying questions here.

joanise:
@dhdaines Rebased and re-opened as a new PR: #258

package.json
dhdaines: the bundler will only bundle stuff that the code actually uses - but als note that the web-component declares the dependencies used by its actual code in `packages/web-component/package.json`

.github/workflows/release.yml
dhdaines: yeah.  my general irritation with nx maybe is showing a bit too much :)

packages/ngx-web-component/package.json
dhdaines: LOL, in fact, the currently released version doesn't actually work outside our monorepo (with angular 15):

```
./src/app/app.module.ts:3:0-72 - Error: Module not found: Error: Can't resolve '@readalongs/web-component/loader' in '/home/dhd/work/ReadAlongs/testng/test-ngx-web-component/src/app'

./node_modules/@readalongs/ngx-web-component/esm2020/lib/ngx-web-component.module.mjs:4:0-72 - Error: Module not found: Error: Can't resolve '@readalongs/web-component/loader' in '/home/dhd/work/ReadAlongs/testng/test-ngx-web-component/node_modules/@readalongs/ngx-web-component/esm2020/lib'

Error: src/app/app.module.ts:4:38 - error TS2307: Cannot find module '@readalongs/web-component/loader' or its corresponding type declarations.

4 import { defineCustomElements } from "@readalongs/web-component/loader";
```

dhdaines: Okay, confirmed that the generated code in `ngx-web-component` is *very* specific to a single version of Angular.  So no it doesn't work with 13 or 14, which is sad really, because there's no good reason for it.

PR #253, opened 2024-04-03
fix: re-enable deprecated useAssetsFolder

### LINKED ISSUE #252: 1.2.0 introduced breaking changes that may cause published read-alongs problems
We just published web-component 1.2.0 with breaking changes, `use-assets-folder="false"` is no longer supported, the equivalent is now `image-assets-folder=""`.

This is an issue because over the last at least one year we generated the snippet

       <script type="module" src='https://unpkg.com/@readalongs/web-component@^1.0.0/dist/web-component/web-component.esm.js'></script>

and `^1.0.0` now resolves to 1.2.0.

Due to the breaking changes, 1.2.0 should probably be yanked (if that's possible) and replaced by 2.0.0?

roedoejet:


joanise (approved):
Great

PR #256, opened 2024-04-03
feat: a WP plugin to load scripts needed for read along web app 

deltork:
This is a plugin that will allow WordPress sites to load the necessary script for the Read Along web app component


packages/web-component/src/components.d.ts
```diff
@@ -6,7 +6,6 @@
  */
 import { HTMLStencilElement, JSXBase } from "@stencil/core/internal";
 import { InterfaceLanguage, ReadAlongMode } from "./index.d";
-export { InterfaceLanguage, ReadAlongMode } from "./index.d";
```
joanise, line 9: AP's recent PR had this line automatically inserted, now this PR automatically removes it. I wonder if we might be working with different versions of node or something. Did you recently redo `npm install` in the project?

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -323,8 +323,45 @@ Please host all assets on your server, include the font and package imports defi
...
+
+        ----- END OF SNIPPET----
+
+        
+        `,
+      ]);
+
       // - add plain text readme
-      innerFolder?.file("README.txt", this.readmeFile);
+      innerFolder?.file("README.txt", WP_deployment_readme);
```
joanise, line 364: I'm having a hard time deciding if I prefer one README.txt having both the regular web deployment guide and the WordPress deployment guide (as you did here), or if I want two separate files. 

But I think this is good as is.

joanise (approved):
This is good. Thanks!

packages/studio-web/src/app/demo/demo.component.ts
deltork: Same with me, so I set it so that we can separate them easily if we want.

PR #257, opened 2024-04-04
[DRAFT] Dev.ej/rebase main onto ng17

joanise:
dev.angular-migration and main have diverged enough that a simple merge is not an option anymore for #247.

This PR is meant to replace #247, by replaying the changes in main since #247 onto its branch.

A merge commit and the `ours` strategy will be required to merge this in, don't let the GitHub UI do the merge! And it won't anyway, because of the conflicts that it can't detect I've fixed here.


joanise:
This was a worthwhile experiment, but it's a dirty solution. I'll leave it here until we review #258 but I don't want to merge this.
It did help me repair the problems I had in #258, though.

PR #258, opened 2024-04-04
Migrate to Angular 17 / Nx 18

### LINKED ISSUE #239: Angular 15 is EOL in May 2024, so bump to Angular 17.

joanise:
Second alternative for rebasing angular migration, and I think this one makes more sense.


joanise:
Question: the readalongs produced by this PR are compatible with web-component v1.2.1, so I think we could revert my version bumping, and let this continue to be called v1.2.1 instead of bumping it to v1.2.2, so that we can finish merging the other changes in progress, e.g., what Del is working on, before making the next minor release.

This would also give us time to rewrite the publication workflow with a tag push, the way we now do in g2p.

dhdaines:
Okay, this should be maybe correct.  Node.js 20.11.0, NPM 10.5.1.

Note that `npm audit fix` basically does not work ever, and this seems to be a well-known problem.

Also, there are some broken dependency specifiers inside this web of 239824513242 packages that mean that you get utterly incomprehensible and unfixable warnings when running `npm audit  fix` even with `--force`.  So we still have some "advisories".

dhdaines (approved):
This seems to be working now!  So okay to merge.

PR #260, opened 2024-04-05
Tonnerre de Brest! Correct the NPM packaging to work outside of this repo!

### LINKED ISSUE #259: @readalongs/ngx-web-component doesn't actually work outside this repo
Trying to recreate the angular-demo standalone:
```
npm install @angular/cli@15
npx ng new test-ngx-web-component
cd test-ngx-web-component
npm install @readalongs/ngx-web-component
```
Oops! That doesn't actualy work, because npm dependency hell:
```
(a bunch of absolute horror npm output snipped)
npm ERR! While resolving: test-ngx-web-component@0.0.0
npm ERR! Found: rxjs@7.8.1
(more horror)
npm ERR! Could not resolve dependency:
npm ERR! peer rxjs@"~7.5.0" from @readalongs/ngx-web-component@1.2.1
(second verse, same as the first, after fixing the rxjs dependency)
npm ERR! While resolving: test-ngx-web-component@0.0.0
npm ERR! Found: zone.js@0.12.0
(etc)
npm ERR! Could not resolve dependency:
npm ERR! peer zone.js@"^0.11.8" from @readalongs/ngx-web-component@1.2.1
```
Well, let's frob package.json in our test app as we have learned to do (probably this can be fixed in web-component...):
```diff
-    "rxjs": "~7.8.0",
+    "rxjs": "~7.5.0",
     "tslib": "^2.3.0",
-    "zone.js": "~0.12.0"
+    "zone.js": "~0.11.8"
```
Okay, now at least we can `npm install`.  So let's copy over the stuff from the angular demo app and see if we can make it work.  First we add the stuff necessary to `src/app/app.module.ts`:
```diff
 import { AppComponent } from './app.component';
+import { NgxRAWebComponentModule } from "@readalongs/ngx-web-component";
+import { defineCustomElements } from "@readalongs/web-component/loader";
+
+defineCustomElements();
 
 @NgModule({
   declarations: [
     AppComponent
   ],
   imports: [
-    BrowserModule
+    BrowserModule,
+    NgxRAWebComponentModule,
```
Does it compile? NO!
```
npx ng serve
(...snip...)
./src/app/app.module.ts:4:0-72 - Error: Module not found: Error: Can't resolve '@readalongs/web-component/loader' in '/home/dhd/work/ReadAlongs/testng15/test-ngx-web-component/src/app'
./node_modules/@readalongs/ngx-web-component/esm2020/lib/ngx-web-component.module.mjs:4:0-72 - Error: Module not found: Error: Can't resolve '@readalongs/web-component/loader' in '/home/dhd/work/ReadAlongs/testng15/test-ngx-web-component/node_modules/@readalongs/ngx-web-component/esm2020/lib'
Error: src/app/app.module.ts:6:38 - error TS2307: Cannot find module '@readalongs/web-component/loader' or its corresponding type declarations.
6 import { defineCustomElements } from "@readalongs/web-component/loader";
```
Strange, since the `@readalongs/web-component` is correctly defined in the `peerDepedencies` and, indeed, `npm` has installed it as expected...

This seems to have something to do with the convoluted way in which the declarations files are structured.  First of all the main types declaration file in `package.json` is `dist/types/components.d.ts` which is almost certainly wrong since there's no possible way you can get to `defineCustomElements` from there.

For this or whatever other reason TypeScript isn't able to resolve "@readalongs/web-component/loader" to `node_modules/@readalongs/web-component/dist/loader/index.d.ts`.  Compounding this is, maybe, it is trying to either import `node_modules/@readalongs/web-component/dist/cjs/loader.cjs.js` or maybe `node_modules/@readalongs/web-component/dist/esm/loader.js` 

The level of complexity in how TypeScript and NPM actually resolve imports is too much for my simple snake brain to handle, so unfortunately I have no proposed solution yet...

dhdaines:
Fixes #259.

Angular 15's `ng new` will give you rxjs `~7.8.0` which is incompatible with the `~7.5.0` we had.  Likewise it gives you zone.js `~0.12.0` which is incompatible with `~0.11.8`.  There's probably no good reason for us to care about these precise versions, but oh well.

More importantly we need to install the `loader` folder alongside `dist`  in the `web-component` for NPM or else we simply can't load the package (we didn't notice this because we were pointing to the built distribution which includes this folder even if the published one to NPM doesn't - I would love to blame this on `nx` but I can't because if we used good old `npm link` we'd have the same problem - NPM is the problem - it is always the problem)

Also make sure the `angular-demo` shows a sea captain.

### DIFF
--- a/packages/web-component/package.json
+++ b/packages/web-component/package.json
@@ -12,7 +12,8 @@
   "unpkg": "dist/bundle.js",
   "private": false,
   "files": [
-    "dist/"
+    "dist/",
+    "loader/"
   ],
   "scripts": {
     "build": "stencil build --docs",
--- a/packages/web-component/package.json
+++ b/packages/web-component/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "audio-recorder-polyfill": "^0.4.1",
     "howler": "^2.2.3",
-    "rxjs": "^6.6.7"
+    "rxjs": ">=6.6.7"
   },
   "devDependencies": {
     "@stencil/angular-output-target": "^0.4.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -14408,9 +14408,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001472",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001472.tgz",
-      "integrity": "sha512-xWC/0+hHHQgj3/vrKYY0AAzeIUgr7L9wlELIcAvZdDUHlhL/kNxMdnQLOSOQfP8R51ZzPhmHdyMkI0MMpmxCfg==",
+      "version": "1.0.30001606",
+      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001606.tgz",
+      "integrity": "sha512-LPbwnW4vfpJId225pwjZJOgX1m9sGfbw/RKJvw/t0QhYOOaTXHvkjVGFGPpvwEzufrjvTlsULnVTxdy4/6cqkg==",
       "dev": true,
       "funding": [
         {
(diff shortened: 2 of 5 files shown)


packages/web-component/package.json
```diff
@@ -12,7 +12,8 @@
   "unpkg": "dist/bundle.js",
   "private": false,
   "files": [
-    "dist/"
+    "dist/",
+    "loader/"
```
roedoejet, line 16: one small change, many headaches 😢

roedoejet (approved):
Looks great - thanks @dhdaines !

PR #261, opened 2024-04-11
fix: remove wordpress info from readme until its ready

roedoejet:


joanise (approved):
OK, we can re-enable this once the documentation is available on how to manually load the WP plugin.

PR #274, opened 2024-04-30
Translations for the Editor, and 3rd party Copyright

joanise:
### PR Goal? 

- Add es & fr messages for the editor
- partially automate the process of updating the es and fr message files
- Add missing 3rd party Copyright notice

Note that this PR seems to change the message file a lot, but it's because I now auto-reorder them to match what `extract-i18n` does, so that I can look at the files in parallel and know where to find the corresponding lines.

Still TODO: re-translate the subtitle message about picking a single-file HTML, once the English is updated.


packages/studio-web/extract-i18n-lang.ts
```diff
@@ -0,0 +1,57 @@
+// Update the language files with the new keys from the English file
+// Usage:
+//   npx tsx extract-i18n-lang.ts src/i18n/messages.json src/i18n/messages.fr.json > src/i18n/messages.fr-updated.json
```
roedoejet, line 3: The README should be update too: https://github.com/ReadAlongs/Studio-Web?tab=readme-ov-file#internationalization-i18n-and-localization-l10n

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -548,7 +548,7 @@ export class ReadAlongComponent {
     if (this.svgOverlay) {
       // either with svg overlay
       this.animateProgressWithOverlay();
-    } else {
+    } else if (play_id) {
```
roedoejet: What's this about? The Cypress tests seem to fail now about something related to scrolling.

roedoejet (changes requested):
Thanks @joanise - looks good, but left a couple comments/questions

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: It's an attempt to fix another bug that's been intermittently popping up in the Cypress tests.

This has been nagging me for a while: every few CI runs, I have to click rerun failed jobs to see them pass. Del and I worked on that yesterday and we thought this would fix it, but apparently not, apparently it just moved the symptom around a bit. We'll have to keep digging I guess.

packages/studio-web/extract-i18n-lang.ts
joanise: Good point.

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: Confirmed, I just clicked "rerun failed jobs" is CI passed this time. I'll take this commit out of this PR, though, since it's not the quick fix I hoped it would be. I'm really looking forward to that bug being addressed!

packages/studio-web/extract-i18n-lang.ts
joanise: done

roedoejet (approved):
thanks so much @joanise !

PR #276, opened 2024-05-01
ci: update publish workflow to work on tag push

### LINKED ISSUE #251: Change the publication workflow to be triggered by a tag push
Mimic g2p's and EveryVoice's new publication workflow and get rid of the release branch, replacing it by the much more standard publish on tag push.

joanise:
### PR Goal? 

Change the publication workflow to work on tag push instead of PR to release

### Feedback sought? 

- how to test this properly
- careful review since it's going to be our publication process

### Priority? 

high

### Tests added? 

n/a

### How to test? 

Good question,  still need to figure this out!

### Confidence? 

medium

### Version change? 

no

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,10 @@ jobs:
           registry-url: "https://registry.npmjs.org"
           scope: "@readalongs"
       - run: npm install
-      - run: npx nx build web-component && npx nx bundle web-component && npx nx build ngx-web-component
+      - run: |
+          npx nx build web-component
+          npx nx bundle web-component
+          npx nx build ngx-web-component
       - run: |
           cd dist/packages/web-component && npm publish --access=public
         env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,6 @@ jobs:
     needs: test
     steps:
       - uses: actions/checkout@v4
-      - uses: fregante/setup-git-user@v2
       # Setup .npmrc file to publish to npm
       - uses: actions/setup-node@v4
         with:
@@ -27,15 +26,21 @@ jobs:
           registry-url: "https://registry.npmjs.org"
           scope: "@readalongs"
       - run: npm install
-      - run: |
+      - name: Build and bundle
+        run: |
           npx nx build web-component
           npx nx bundle web-component
           npx nx build ngx-web-component
-      - run: |
+      - name: Check that the web-component bundle was up to date
+        run: |
+          if git status --porcelain | grep packages/studio-web/src/assets/; then echo ERROR: The web-component bundle is out of date. Please run \"npx nx bundle web-component\" and commit the results, then update the tag and try again.; false; fi
+      - name: Publish web-component to npmjs
+        run: |
           cd dist/packages/web-component && npm publish --access=public
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-      - run: |
+      - name: Publish ngx-web-component to npmjs
+        run: |
           cd dist/packages/ngx-web-component && npm publish --access=public
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,18 +1,25 @@
-# This workflows will build, test and upload the package to npm when a release is created
+# This workflows will build, test and upload the package to npm when a release
+# is created.
+# To create a release, please follow the steps documented in the "Build &
+# Publish" section of the README.md file.
 
 name: Upload Package
 on:
   push:
-    branches: [release]
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+
 jobs:
+  test:
+    uses: ./.github/workflows/end-to-end-tests.yml
+    secrets: inherit
+
   build:
     runs-on: ubuntu-20.04
+    needs: test
     steps:
       - uses: actions/checkout@v4
-      - uses: fregante/setup-git-user@v1
+      - uses: fregante/setup-git-user@v2
       # Setup .npmrc file to publish to npm
       - uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -44,33 +36,17 @@ jobs:
...
-        id: determine_tag
-        run: |
-          echo "TAG_VERSION=$(npm view node_modules/@readalongs/web-component version)" >> $GITHUB_OUTPUT
-      - name: Bump version and push tag
-        id: tag_version
-        uses: mathieudutour/github-tag-action@v6.1
+      - name: Update CHANGELOG
+        id: changelog
+        uses: requarks/changelog-action@v1
         with:
```
roedoejet, line 52: it might be nice to have a check here that determines whether npm has the version available (i.e. whether we forgot to bump). `npm view @readalongs/web-component version` should display the most recent version, and that should be less than the current defined version in the code/tag

roedoejet: if it doesn't, we shouldn't create a github release, and this should fail.

README.md
```diff
@@ -185,27 +185,30 @@ or
...
-`@readalongs/web-component` in
-`packages/ngx-web-component/package.json` - if you find this to be
-less than useful feel free to add your voice to [this bug
-report](https://github.com/nrwl/nx/issues/19989)), run `npm install`
-to reflect that bump in `package-lock.json`, and build both the web
-component and angular wrapper:
+NPM. Then, apply all the version bump and `npm install` steps above,
+and build both the web component and angular wrapper:
 
     npx nx build web-component
```
roedoejet, line 214: Don't we need to also run `npx nx bundle web-component`?

roedoejet (changes requested):
Looks great @joanise - and super quickly done 👏 

I've made a request for something that checks that we bumped the version, and also a clarifying question for the manual publishing.

joanise:
Tested the workflow by creating a variant that ran on push to `test1.2.3` and did dry run operations.
See https://github.com/ReadAlongs/Studio-Web/actions/runs/8914132259/job/24481166770
It looks like it works fine.

README.md
joanise: That's just a few lines lower in the README.md. It's needed before running `npx nx build studio-web`, but since we don't publish studio-web it's not that relevant...

.github/workflows/release.yml
joanise: Will `npm publish` fail if you try to republish the same version again?

joanise: But good point, we should make sure the GH release doesn't happen in this case.

joanise: Indeed, we should already be safe: https://docs.npmjs.com/cli/v8/commands/npm-publish says:
> The publish will fail if the package name and version combination already exists in the specified registry.

and once the first publish fails the entire workflow will abort.

README.md
joanise: - Updated the readme to say you need to do this before even creating the tag.
 - Updated the release workflow to fail early if the bundle is out of date.

.github/workflows/release.yml
roedoejet: hmm I feel like if that's the case, it's new. I seem to remember things not aborting just because npm publish failed

.github/workflows/release.yml
```diff
@@ -1,76 +1,60 @@
...
-      - run: |
-          export VERSION=$(npm view node_modules/@readalongs/web-component version)
-          perl -i -lne 'if (/peerDependencies/) {print; print "\"\@readalongs/web-component\": \"'$VERSION'\","} else { print }' dist/packages/ngx-web-component/package.json
-      - run: |
+      - name: Build and bundle
+        run: |
+          npx nx build web-component
+          npx nx bundle web-component
+          npx nx build ngx-web-component
+      - name: Check that the web-component bundle was up to date
```
dhdaines, line 36: We could put a check here to make sure the dependency in ngx-web-component is correct as well, since we're asking people to do it manually in the README.md..

README.md
```diff
@@ -185,27 +185,31 @@ or
...
+#### Preparing to publish the Web Component & Angular Wrapper
 
-The publication of the web component and its angular wrapper has been automated in the `release.yml` workflow. To trigger it, submit a pull request to the `release` branch and get it reviewed. Publication will happen when the PR is merged.
+Publication to npmjs is mostly automated via the tag-triggered `release.yml` CI workflow, but first you have to update the repo with the new version number in multiple places:
 
-#### Web Component & Angular Wrapper - manually
+- Update the version number for web-component and ngx-web-component in [packages/web-component/package.json](packages/web-component/package.json) and [packages/ngx-web-component/package.json](packages/ngx-web-component/package.json).
+- Nx does not manage dependencies among projects for us, so update the version in `peerDependencies` for `@readalongs/web-component` in [packages/ngx-web-component/package.json](packages/ngx-web-component/package.json).
+- Update `singleFileBundleVersion` in [packages/studio-web/package.json](packages/studio-web/package.json).
+- All the versions above should be exactly in sync!
```
dhdaines, line 195: As mentioned above... we should verify this in the release workflow and fail if not.  I will try adding some code to do this shortly.

dhdaines (changes requested):

.github/workflows/release.yml
dhdaines: Added this in 2604892

dhdaines (approved):
Okay, now I approve, with the change I added to verify the versions.  I tested it on the command-line locally, unsure how to test it in actual CI but it seems correct.

joanise:
Darn, @dhdaines we worked in parallel. Before leaving work yesterday, I had almost completed the same checks you did later in the evening. See https://github.com/ReadAlongs/Studio-Web/commit/decc1d72b2faf0168820fc9404bf953e5a51a652 
My checks are a little more complete, but use jq instead of npm. Do you have a preference between your solution and mine?

dhdaines:
> Darn, @dhdaines we worked in parallel. Before leaving work yesterday, I had almost completed the same checks you did later in the evening. See [decc1d7](https://github.com/ReadAlongs/Studio-Web/commit/decc1d72b2faf0168820fc9404bf953e5a51a652) My checks are a little more complete, but use jq instead of npm. Do you have a preference between your solution and mine?

Yours are more complete but I think it would be preferable to use `npm` since it's already there... should be pretty easy to just drop in the `npm` calls that I did.

joanise:
> Yours are more complete but I think it would be preferable to use `npm` since it's already there... should be pretty easy to just drop in the `npm` calls that I did.

`jq` is there by default on all the runner images on GitHub, so being there is not an issue. I like how `jq` can run earlier, before we even build, and thus fail early. But I like how your `npm` commands run on actual builds in `dist/`. (And, ah-ha! that's how to do it locally! I tried and failed to get `npm view` to dome something useful with `packages/web-component`.)

I don't see an `npm view` equivalent for `"$(jq -r .singleFileBundleVersion packages/studio-web/package.json)" != "$VERSION"`, though, since we don't build or publish that package, but we still want to check this.

dhdaines:
Ah, good to know ... so no problem to use `jq`, or `npm view` either for that matter :) your checks are more complete and probably more correct.

joanise:
> Ah, good to know ... so no problem to use `jq`, or `npm view` either for that matter :) your checks are more complete and probably more correct.

Thanks for doing yours, though, and sorry for not flagging that I was writing them yesterday.

roedoejet (approved):

PR #283, opened 2024-05-08
fix(web-c): better Fr trans for "upload an image for this page"

joanise:
"Téléverser charger" was just wrong, but I completely rewrote the message to be clearer and closer to what it means rather than to the Enlish text it was translated from.

### PR Goal? 

Tiny translation fix in the French

### Priority? 

low

### Tests added? 

n/a

### How to test? 

choose the Web-C language as French 

### Confidence? 

high

### Version change? 

not just for this, but it'll come with the minor bump already scheduled.


dhdaines (approved):
Oh, oops, yes, this is much better!

PR #286, opened 2024-05-09
fix: allows the ability to highlight the word using the progress bar

### LINKED ISSUE #266: re-align "highlighted" word / page  focus when  moving forward / backwards without needing to "play"
Currently if the RA is "stoped"  and you click the "5 second" back button  , or move/click on the timeline   the screen will only "re-align"  to the nearest word when you click on the play button.        But  by the time the pages re-aligns  ( page transition  for example)  , the user will probably miss the first few words spoken.

It would be nice if RA would re-align  without needing to  'play' after moving the timeline.

### LINKED ISSUE #275: CI randomly fails in basic.cy.js, apparently related to scrolling
Intermittent CI workflow failure, so this is hard to reproduce. But see
https://github.com/ReadAlongs/Studio-Web/actions/runs/8903159213/job/24450409271
https://github.com/ReadAlongs/Studio-Web/actions/runs/8900297255/job/24441553848
https://github.com/ReadAlongs/Studio-Web/actions/runs/8885442422/job/24396804150

Now, the patch 
```
-    } else {
+    } else if (play_id) {
```
to `read-along.tsx` modified the error message but unfortunately did not actually make the problem go away.

deltork:
### PR Goal? 

- Allows the ability to highlight the words in the read-along using the progress bar, even if the audio is not currently playing
- Changes to the web-component tests to use a new state-sensitive custom command for play sequence. 

### Fixes? 
Closes #266  #275 

### Feedback sought? 
Progress bar interaction, play/stop.

### Priority? 
Low

### Tests added? 
Updates all web-components tests.

### How to test? 
Clicking various locations on the progress bar, start/pause/stop a bunch of time 

### Confidence? 
high

### Version change? 

no

### DIFF
--- a/packages/web-component/cypress/support/commands.js
+++ b/packages/web-component/cypress/support/commands.js
@@ -28,3 +28,9 @@ Cypress.Commands.add("readalong", () => {
 Cypress.Commands.add("readalongElement", () => {
   return cy.get("read-along").first();
 });
+
+Cypress.Commands.add("playReadAlong", () => {
+  cy.get("[data-cy=play-button]", { timeout: 10000 })
+    .should("be.enabled")
+    .click();
+});
--- a/packages/web-component/src/components/read-along-component/read-along.tsx
+++ b/packages/web-component/src/components/read-along-component/read-along.tsx
@@ -450,7 +450,7 @@ export class ReadAlongComponent {
    */
   play() {
     //do not attempt to play if sprites are not initialized
-    if (this.audio_howl_sprites == undefined) return;
+    if (this.audio_howl_sprites === undefined) return;
     this.playing = true;
     // If already playing once, continue playing
     if (this.play_id !== undefined) {
@@ -1122,7 +1122,7 @@ export class ReadAlongComponent {
         }
         // Turn tag to query
         let query = this.tagToQuery(el_tag);
-        if (query == undefined) return; // not go any further if tag does not exist in the DOM
+        if (query === undefined) return; // not go any further if tag does not exist in the DOM
         // select the element with that tag
         let query_el: HTMLElement = this.el.shadowRoot.querySelector(query);
         // Remove all elements with reading class
@@ -1215,7 +1215,7 @@ export class ReadAlongComponent {
     if (this.audio_howl_sprites === undefined) return false;
 
     //not ready if
-    if (this.audio_howl_sprites.sound == undefined) return false;
+    if (this.audio_howl_sprites.sound === undefined) return false;
 
     return true; //it is ready
   }
--- a/packages/web-component/src/components/read-along-component/read-along.tsx
+++ b/packages/web-component/src/components/read-along-component/read-along.tsx
@@ -10,6 +10,7 @@ import {
   Method,
   Prop,
   State,
+  Watch,
 } from "@stencil/core";
 
 import {
@@ -201,18 +202,20 @@ export class ReadAlongComponent {
   finalTaggedWord: string;
   @State() settingsVisible: boolean = false;
   @State() userPreferencesDirty: boolean = false;
+  @Watch("audio_howl_sprites")
   /************
    *  LISTENERS  *
    ************/
-
   @Listen("wheel", { target: "window" })
   wheelHandler(event: MouseEvent): void {
     // only show guide if there is an actual highlighted element
     if (this.el.shadowRoot.querySelector(".reading")) {
       if (
-        event["path"][0].classList.contains("sentence__word") ||
-        event["path"][0].classList.contains("sentence__container") ||
-        event["path"][0].classList.contains("sentence")
+        event["path"] &&
+        event["path"].length > 0 &&
+        (event["path"][0].classList.contains("sentence__word") ||
+          event["path"][0].classList.contains("sentence__container") ||
+          event["path"][0].classList.contains("sentence"))
       ) {
         if (this.autoScroll) {
           let reading_el: HTMLElement =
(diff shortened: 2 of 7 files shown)


packages/web-component/cypress/support/commands.js
```diff
@@ -28,3 +28,7 @@ Cypress.Commands.add("readalong", () => {
 Cypress.Commands.add("readalongElement", () => {
   return cy.get("read-along").first();
 });
+
+Cypress.Commands.add("playReadAlong", () => {
+  cy.get("[data-cy=play-button]").should("be.enabled").click();
```
joanise: Ah, clever, this means -- I assume -- we'll wait until it's enabled before clicking it? That makes sense, so this means sometimes were were "clicking" on it before it was even enabled!

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1706,7 +1713,11 @@ export class ReadAlongComponent {
   PlayControl = (): Element => (
     <button
       data-cy="play-button"
-      disabled={this.hasLoaded < 2}
+      disabled={
+        this.hasLoaded < 2 &&
+        this.audio_howl_sprites &&
+        this.audio_howl_sprites.sound?.duration() > 0
```
joanise: What does `sound?.duration()` return when `sound` is undefined? Will that return value be comparable via `> 0`? I assume the intent is if sound is not defined, you want `> 0` to be false, but I just want to make sure we don't raise an exception trying to evaluate `undefined > 0` instead of just having it be fast.

joanise: The logic here does not make sense to me, though: we're saying 

    disabled = loaded < 2 && sprites && duration > 0

isn't this backwards? I.e., shouldn't it be

    enabled = loaded >= 2 && sprites && duration > 0
    disabled = !enabled

or maybe more directly

    disabled = loaded < 2 || !sprites || !duration

and maybe even actually 

```
      disabled={
        this.hasLoaded < 2 ||
        !this.audio_howl_sprites?.sound?.duration()
      }
```

joanise: And finally, why is the definition of `disabled` in curly braces? Why not regular round ones?
(That's not your change, it was already like that, but I don't understand it.)

packages/web-component/src/utils/utils.ts
```diff
@@ -186,15 +186,25 @@ Sprite.prototype = {
     var seek = self.sound.seek((id = id));
     for (var j = 0; j < self._spriteLeft.length; j++) {
       // if seek passes sprite start point, replace self._reading with that sprite and slice the array of sprites left
-      if (seek * 1000 >= self._spriteLeft[j][0]) {
+      if (seek >= self._spriteLeft[j][0]) {
```
joanise, line 187: was this a ms vs s mistake all along?

joanise (changes requested):
Tested, it works very well. Some comments, but most importantly a question about the semantics of the `disabled` variable.

packages/web-component/src/components/read-along-component/read-along.tsx
deltork: Created a function to walk through the states required for the successful playing of RA

packages/web-component/cypress/support/commands.js
deltork: Yes!

packages/web-component/src/utils/utils.ts
deltork: Good catch. This is a result of proper conflict resolution when rebasing. I readded the conversion to ms in the line (184) above.

deltork (commented):
As discussed I have run the CI (see logs https://github.com/ReadAlongs/Studio-Web/actions/runs/9102062098/job/25055245809) 10 times and it has passed 10 times. This patch is ready.

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -443,6 +449,8 @@ export class ReadAlongComponent {
    *
    */
   play() {
+    //do not attempt to play if sprites are not initialized
+    if (this.audio_howl_sprites == undefined) return;
```
joanise: I believe this should use `===`.

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1187,7 +1201,24 @@ export class ReadAlongComponent {
...
+   * This could be expressed as one line
+   * but because there are multiple states involved
+   * breaking it down to make it easier to track
+   */
+  isReadyToPlay(): boolean {
+    //not ready if still audio & alignment is not loaded
+    if (this.hasLoaded < 2) return false;
 
+    //not ready if
+    if (this.audio_howl_sprites == undefined) return false;
```
joanise: ===

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1187,7 +1201,24 @@ export class ReadAlongComponent {
...
+
+    //not ready if
+    if (this.audio_howl_sprites.sound == undefined) return false;
```
joanise: ===

joanise (changes requested):
Now just a trivial change, my understanding is we should always test for undefined using `===` rather than `==`, otherwise I think this is ready to merge. The web-component works really nicely with these changes.

joanise (approved):
I made the other changes I requested, this is ready now.
Thanks for fixing this problem, I'll be happy to have stable CI back!
And the interface is nicer to use now, which is great!

PR #292, opened 2024-05-15
ci: delete PR previews when the PR is closed

### LINKED ISSUE #285: Clean up PR previews once we're done with them
Right now, PR previews accumulate without limit, and I have twice manually removed old ones.

The problem: each PR preview includes three copies of the en-us model from SoundSwallower, and that makes our gh-pages artefacts quite large (586 MB before I cleaned up today, 258MB now).

I tried replacing all the copies with symlinks, but that turns out not to be a solution: the upload artifact action dereferences the symlinks as it creates the artifact! ☹️ 

On https://github.com/marketplace/actions/deploy-pr-preview, the authors of the PR preview action actually recommend running on PR `closed` events too, and it will automatically delete previews for closed PRs for us. I think this is what we should do, to keep our gh-pages deployments small, as recommended.

We're still clear of the 1GB hard limit at the moment, but I believe we would have busted it if I had not cleaned up manually. I'd like to use a self-cleaning setup so I don't have to think about it.

joanise:
### PR Goal? 

Auto-clean up PR previews

### Fixes? 

Fixes #285

### Feedback sought? 

rubber stamping

### Priority? 

normal

### Tests added? 

n/a, the test will be looking at what happens when we merge this PR in.

### How to test? 

merge the PR. Kinda late, but it's a trivial PR

### Confidence? 

high, I followed the simple instructions at https://github.com/marketplace/actions/deploy-pr-preview

### Version change? 

no

### DIFF
--- a/.github/workflows/pr-preview.yml
+++ b/.github/workflows/pr-preview.yml
@@ -7,6 +7,7 @@ on:
       - opened
       - reopened
       - synchronize
+      - closed
 jobs:
   deploy-preview:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
Nice! I haven't seen this syntax before. I assume it's been tested out in some way?

PR #306, opened 2024-06-03
fix: update unidecode

roedoejet:
fixed by https://github.com/FGRibreau/node-unidecode/issues/31

### PR Goal? 

We used to have an issue with some characters being unidecoded as ?

### Fixes? 

This updates our dependency on unidecode to [the new version](https://github.com/FGRibreau/node-unidecode/issues/31) that fixes that.

### Feedback sought? 

sanity

### Priority? 

medium

### Tests added? 

none

### How to test? 

we slugify anyways, so it shouldn't really matter (other than improving the filenames a bit more)

### Confidence? 

high (I'm just updating the dependency here)

### Version change? 

no

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "shepherd.js": "^11.2.0",
         "soundswallower": "^0.6.3",
         "standardized-audio-context": "^25.3.70",
-        "unidecode": "^1.0.1",
+        "unidecode": "^1.1.0",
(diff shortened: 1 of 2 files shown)


joanise:
Nice, SȾÁ,SEN now downloads as `sta-sen` instead of `s-a-sen`.

joanise (approved):
Fantastic that this got fixed. We'll get nicer filenames.

PR #317, opened 2024-06-18
Apply a few of the misc updates that were assigned

### LINKED ISSUE #314: In the web bundle, rename the editable file to make it more self-documenting and intuitive
Based on user feedback, "editable" is confusing. Indeed, the format is pretty opaque, while the multi file format is much more manually editable. What we mean is compatible with our editor, but that's not clear.

Desiderata
 - since it's the same file you get when you pick Offline HTML as the download option, make its name consistent with that (e.g., maybe `Offline-HTML/slugified-title.html`?)
  - make the connection to #299 point about "Editor interface: make it clearer where the file to upload comes from"

roedoejet:
### PR Goal? 

Add improved filelist names to the web bundle and more information to the tour

### Fixes? 

https://github.com/ReadAlongs/Studio-Web/issues/314
https://github.com/ReadAlongs/Studio-Web/issues/303

### Feedback sought? 

Sanity check.

*Note*: I will need i18n updates to these commits @joanise - merci!

### Priority? 

low

### Tests added? 

none (just changing names of files basically)

### How to test? 

Run the tour, or try downloading the web bundle

### Confidence? 

high - tested on my machine.

### Version change? 

No


packages/studio-web/src/app/editor/editor.component.html
```diff
@@ -23,7 +23,9 @@ <h1 i18n="Welcome message for app editor" id="welcome-header">
     <div class="row mb-6">
       <div class="col-8">
         <h2 i18n="Title for upload box" class="subtitle">
-          Select a single-file HTML ReadAlong file
+          Select an offline HTML ReadAlong file. To get this file, you should
+          have created a ReadAlong using the <a routerLink="/">Studio</a>, and
+          then selected "Offline HTML" as the output.
```
joanise: add "or use the Offline-HTML file included in the Web Bundle download format" or something like that.

joanise (changes requested):
Wow, that was fast. I assume you'd like me to do the translations. I have a small text change suggestion first, then I'll  put that on my TODO list.

deltork (approved):
Everything looks good! I am approving pending i18n fix. Thanks for taking care of this

joanise (approved):
Fixed done, translations added, ready to merge.

PR #323, opened 2024-07-11
Dev.del/navbar-patch

deltork:
patch to fix version display issue in editor


joanise:
With Aidan's version, I get a horizontal scrollbar on opening, with a tiny white bar off screen on the right. Oh, and same in live deployment, so that's not new.
With this version, I get both a horizontal and a vertical scrollbar on opening, and the version is hidden by the scrollbar so I have to scroll down to see it. Each scroll bar is there because the default view is a tiny bit bigger than the viewport.
It would be nice to get rid of both of these!

deltork:
adding this change to branch dev.del/editor-tour

PR #324, opened 2024-07-16
docs: add quick links to the editor and the documentation

joanise:
### PR Goal? 

Minimal link option pointing to Editor and docs, not the way we will want them for good, but while we want for Aidan's better solution to be fixed.

I suggest merging this once Del has fixed the download output from the editor, if you're OK with this format.

### DIFF
--- a/packages/studio-web/src/app/studio/studio.component.html
+++ b/packages/studio-web/src/app/studio/studio.component.html
@@ -14,18 +14,44 @@
           <h1 i18n="Welcome message for app" id="welcome-header">
             Welcome to ReadAlong Studio
           </h1>
-          <p i18n="Description of app">
-            This is a tool to help you make your own interactive 'readalong'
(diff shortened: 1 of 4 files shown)


joanise:
@deltork This should only be merged once you feel it's OK to make the editor visible to the world. After #325 is fixed and enough sanity check that you're happy with it.

deltork (approved):

roedoejet (changes requested):
Hm, I think this is an improvement from before the navbar was added, but I think the link to the editor is redundant given the new navbar. I also think the documentation link should be more hidden or maybe not even included since we actually want folks to go on the tour, not navigate away to the documentation. I'm going to close this for now instead of merging with the navbar code (which touches the same stuff, and would require a more conscious merge that I don't think we can do while @joanise is away). I'll close this for now, but happy to revisit when @joanise gets back

PR #326, opened 2024-07-20
fix state in studio and editor

roedoejet:
### PR Goal? 

refactor all the stateful logic from the studio and editor into services

### Fixes? 

The issue brought up in https://github.com/ReadAlongs/Studio-Web/pull/322

### Feedback sought? 

- Sanity checks
-  help identifying how to test properly

### Priority? 

high - I think this touches enough things that we should merge before starting new work

### Tests added? 

none yet

### How to test? 

try creating a readalong or editing one and then navigate away from the studio or editor. When you return, your content (and any changes) should still be there

### Confidence? 

medium-low

### Version change? 

no - only touches the studio


deltork (approved):
Looks good. 

State persists across tabs; you get a warning when navigating away from the page.

Thanks for doing this

Issue #333, opened 2024-08-27
The download format option disappears on tablet-size screens and smaller


joanise:
When the viewport width is <=992, the Output Format selector next to the download button, in both the Editor and the Studio, disappears and the user can only download single-file HTML.

deltork:
see PR #334 
![image](https://github.com/user-attachments/assets/f34e50b9-cf89-42d7-a07e-7e656737c78d)
![image](https://github.com/user-attachments/assets/ef9aad82-14a0-4fe4-a525-db4c53068a2f)

PR #334, opened 2024-09-09
fix: #333 made the studio step two header, mobile friendly

deltork:
### PR Goal? 
fix a bug

### Fixes? 
#333

### Feedback sought? 
sanity check

### Priority? 
high

### Tests added? 
n/a

### How to test? 
use the browser device emulator on the PR preview

### Confidence? 
very

### Version change? 
patch

### DIFF
--- a/packages/studio-web/src/app/demo/demo.component.html
+++ b/packages/studio-web/src/app/demo/demo.component.html
@@ -1,7 +1,7 @@
 <section>
   <div class="container" *ngIf="studioService.render$ | async">
     <div class="row">
-      <div class="col-9">
+      <div class="col col-md-6 col-lg-9">
         <h2
           id="readalong-congratulations"
           i18n="Congratulations message"
@@ -10,7 +10,7 @@
           Congratulations! Here's your ReadAlong!
         </h2>
       </div>
-      <div class="col-3">
+      <div class="col col-md-6 col-lg-3">
         <ras-shared-download
           (downloadButtonClicked)="download($event)"
         ></ras-shared-download>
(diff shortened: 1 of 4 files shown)


dhdaines:
Hmm. Doesn't seem to be very iPad friendly (let alone phone friendly) according to Chrome's dev tools:
image

deltork:
> Hmm. Doesn't seem to be very iPad friendly (let alone phone friendly) according to Chrome's dev tools: image

this is what it looks like Chrome dev tools on windows 11:
image


this is what is looks like Edge dev tools on windows 11: 
image

dhdaines:
Hmm, maybe I'm not picking up the right version from the dev preview?  I'll try rebuilding locally to see.

deltork:
> Hmm, maybe I'm not picking up the right version from the dev preview? I'll try rebuilding locally to see.

It might be, I used the PR preview https://readalong-studio.mothertongues.org/pr-preview/pr-334/#/ for my screenshot 
image

dhdaines:
I think there's just something weird going on with responsive design mode on my Chrome, because it looks fine in a normal window when I resize it...

joanise:
This looks quite good to me, thanks! On a small screen, the download button ends up taking more real estate that I hoped.

I imagined it shrinking the format list, rather than wrapping it, something like a little triangle that when you clicked on it, it expanded the list, but otherwise you'd see just this triangle left of the download arrow if there's no space for the long drop down.

But I'm scope creeping here, I'm happy to approve as is as well.

joanise (approved):
lgtm

PR #344, opened 2024-09-24
build: make bundle compatible with Windows and test it in CI

joanise:
In CI, let's this npx nx bundle web-component on both Windows and Linux, and on Linux let's display the diff in a way that we can see the bits that changed even though it's minified.

### PR Goal? 

Make `npx nx bundle web-component` compatible with Windows, and exercise it in CI

### Fixes? 

the fact that the bundle command was specifically a bash command, but did not enforce being rush by bash

### Feedback sought? 

rubber stamping

### Priority? 

low

### Tests added? 

yes

### How to test? 

run `npx nx bundle web-component` on Windows and see it no longer fail

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/packages/web-component/bundle.sh
+++ b/packages/web-component/bundle.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# This script is used to bundle the web component for including in Offline HTML files
+# created by studio-web
+
+set -o errexit
+set -o verbose
+
+node b64Fonts.js
+npx webpack --config webpack.config.js
+packageVersion=$(npm view ../../node_modules/@readalongs/web-component version)
+timestamp=$(date "+%Y-%m-%d+%H-%M-%S")
+cd ../studio-web
+npm pkg set singleFileBundleVersion="${packageVersion}"
+npm pkg set singleFileBundleTimestamp="${timestamp}"
--- a/packages/web-component/package.json
+++ b/packages/web-component/package.json
@@ -15,7 +15,7 @@
     "dist/"
   ],
   "scripts": {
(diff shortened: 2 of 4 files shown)


packages/web-component/bundle.sh
```diff
@@ -0,0 +1,15 @@
...
+# This script is used to bundle the web component for including in Offline HTML files
+# created by studio-web
+
+set -o errexit
+set -o verbose
+
+node b64Fonts.js
+npx webpack --config webpack.config.js
+packageVersion=$(npm view ../../node_modules/@readalongs/web-component version)
+timestamp=$(date "+%Y-%m-%d+%H-%M-%S")
```
dhdaines, line 12: Guessing that if you have `bash` on Windows then you also have `date`?

packages/web-component/package.json
```diff
@@ -15,7 +15,7 @@
     "dist/"
   ],
   "scripts": {
-    "bundle": "node b64Fonts.js && webpack --config webpack.config.js; packageVersion=$(npm view ../../node_modules/@readalongs/web-component version); timestamp=$(date \"+%Y-%m-%d+%H-%M-%S\") && cd ../studio-web && npm pkg set singleFileBundleVersion=\"${packageVersion}\" && npm pkg set singleFileBundleTimestamp=\"${timestamp}\"",
+    "bundle": "bash bundle.sh",
```
dhdaines, line 18: well, that's ... much nicer looking

dhdaines (approved):
Good idea!

packages/web-component/bundle.sh
joanise: Yeah, once you're in bash, you get the standard *nix command line tools, even on Windows, through msys, mingw64, cygwin or some derivative. I'm not aware of any instance of bash on Windows that's not based on one of those. Git systematically comes with Git Bash on Windows, that's the one I always use now, it's lightweight but has everything I need to make my Windows machine pleasant to use.

PR #348, opened 2024-10-03
fix: always auto-update the browserslist db on deploy

joanise:
### PR Goal? 

When we deploy, I often notice this message in the logs:

```
Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
```

Instead of trying to remember to do it periodically, it seems wise to just automate it.

This automation here is to update the db and commit the changes when we are testing on a deb branch (i.e., a non-protected branch).

I also fixed some concurrency problems that I noticed while testing this PR.

### Priority? 

low

### Confidence? 

high

### DIFF
--- a/.github/workflows/dev-preview.yml
+++ b/.github/workflows/dev-preview.yml
@@ -1,5 +1,7 @@
 name: Deploy dev preview
-concurrency: dev-preview
+concurrency:
+  group: dev-preview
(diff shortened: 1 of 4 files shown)


dhdaines (approved):
seems good!

PR #349, opened 2024-10-03
Dev.ej/translate route buttons

joanise:
### PR Goal? 

Right now, the Studio / Editor / Privacy buttons on the top right of the Studio and Editor are shown in English on the French and Spanish versions. This PR wants to translate them.

### Feedback sought? 

This should be rubber stamping, but the longer words for Privacy (Privacidad and, much worse, Confidentialité) make the buttons too wide. E.g., on iPhone SE:
fr image
es image
vs English image

At a more reasonable screen size, in English, you fully see the buttons down to 437 pixels wide, and that's perfectly acceptable, but in French you need at least 500 pixels wide to see "Confidentialité" in full. Is there a way to have each of the three buttons auto-shrink when they stop fitting? Ideally, each of the three words would be partially trimmed on the right so that you still see some of each button's text even on a narrow cell phone with 375 pixels wide.

### Priority? 

medium

### How to test? 

load the PR preview in each language and play with simulated narrow screens

### Confidence? 

still low because of the size problems

### Version change? 

no


joanise:
Rebased this PR to the reverted main branch so it can be reviewed properly. It's on top of other PRs, but should still be reviewed by itself.

joanise:
And it still needs a solution for seeing all three buttons when the text is wider, in French and Spanish, as I initially commented.

roedoejet:
Yes, there's a way to do that. at a minimum we could put `overflow: scroll;` on the toolbar which would allow people to scroll.   However I think maybe a better thing to do would be to collapse the toolbar into a ☰ icon for small viewports which expands a menu when clicked. I.e. https://material.angular.io/components/menu/overview. I've made a PR [here](https://github.com/ReadAlongs/Studio-Web/pull/361) for this.

joanise:
Thanks @roedoejet for this suggestion. I've merged it here, with some improvements Del suggested.

I think this PR is now ready to review, and hopefully to merge, although I'll clean up the commit history before I do so.

packages/studio-web/project.json
```diff
@@ -49,8 +49,8 @@
           "budgets": [
             {
               "type": "initial",
-              "maximumWarning": "500kb",
-              "maximumError": "2mb"
+              "maximumWarning": "1500kb",
+              "maximumError": "3mb"
```
joanise, line 53: We were busting our 2mb budget deploying this, I think it's reasonable to go to 3mb, that's still reasonably small.

packages/studio-web/src/app/editor/editor.component.html
```diff
@@ -6,7 +6,7 @@
         <h1 i18n="Welcome message for app editor" id="welcome-header">
           Welcome to the ReadAlong Studio Editor
         </h1>
-        <p>
+        <p i18n="Editor synopsis">
```
joanise, line 9: Only loosely related to the rest of this PR: I noticed this text was still in English on the fr and es versions.

deltork (commented):
Looks good!

deltork (approved):

PR #351, opened 2024-10-10
fix: cannot download other formats

### LINKED ISSUE #347: Cannot download Elan, Praat, SRT or WebVTT from Editor
If you create a read-along in the Studio, all six download formats work fine.

If you then reload any read-along in the Editor, html and zip downloads work, but the other four formats give a 422 with this error:

```
{
    "detail": "ReadAlong provided is not valid: <string>:1:0:ERROR:VALID:DTD_UNKNOWN_ATTRIBUTE: No declaration for attribute xmlns of element read-along"
}
```
as seen by going F12 / Network / <pick download event with the 422> / Response

deltork:
### PR Goal? 
bug fix

### Fixes? 
#347 

### Feedback sought? 
sanity check

### Priority? 
high

### Tests added? 
to be added to studio test suites

### How to test? 
use the pr-preview editor to upload an offline format and try to export non web formats

### Version change? 
bug fix

This PR depends on the merging of PR https://github.com/ReadAlongs/Studio/pull/247

### DIFF
--- a/packages/studio-web/src/app/editor/editor.component.ts
+++ b/packages/studio-web/src/app/editor/editor.component.ts
@@ -252,9 +252,9 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
 
   async parseReadalong(text: string): Promise<string | undefined> {
     const parser = new DOMParser();
+
     const readalong = parser.parseFromString(text, "text/html");
     const element = readalong.querySelector("read-along");
-
     if (element === undefined || element === null) {
       return undefined;
     }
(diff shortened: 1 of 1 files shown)


joanise:
This looks safe to merge already, even before we merge https://github.com/ReadAlongs/Studio/pull/247
I'm not sure why it's labelled draft, it seems to work correctly already. Given it would fix the live deployment, I'm tempted to merge it ASAP, @roedoejet if you have a chance to review it too maybe we could do that soon, even while Del is away.

joanise (approved):
Looks good to me.

roedoejet:
> This looks safe to merge already, even before we merge [ReadAlongs/Studio#247](https://github.com/ReadAlongs/Studio/pull/247) I'm not sure why it's labelled draft, it seems to work correctly already. Given it would fix the live deployment, I'm tempted to merge it ASAP, @roedoejet if you have a chance to review it too maybe we could do that soon, even while Del is away.

I don't understand parts of this PR, but I unfortunately don't have enough time to dig into it either. Del is away for the weekend? If you're confident in these changes, go for it, @joanise otherwise my in-depth review would have to wait until next week

joanise:
> I don't understand parts of this PR, but I unfortunately don't have enough time to dig into it either. Del is away for the weekend? If you're confident in these changes, go for it, @joanise otherwise my in-depth review would have to wait until next week

Fair enough. I guess I'll test more, to make sure it really does work as I think it does, and decide what to do based on that.

packages/studio-web/src/app/editor/editor.component.ts
```diff
@@ -263,16 +263,20 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
...
+    const textNode = element.querySelector("text");
+    if (textNode?.querySelector("body") == null) {
+      while (textNode?.hasChildNodes()) {
         // @ts-ignore
         body.appendChild(textNode.firstChild);
       }
-      textNode.appendChild(body);
-    }
+      textNode?.appendChild(body);
+    } /**/
```
joanise, line 273: I have to say I don't understand how this code works, but I tested it carefully, and what I can see is that it does work. I took a readalong, changed its `body` tag to something else in the base64-encoded .readalong file, and this code now correctly insert a new `body` tag in the right place. Whereas the old code did not, and causes trouble as we know.

joanise (commented):
Alright, carefully tested, I'm going to merge this now, deploy it, and even release 1.5.1 with it because I think it's an important and urgent patch.

PR #355, opened 2024-10-17
fix: cannot download other formats

joanise:
Re-apply #351 on reverted main branch.


joanise:
neved mind, this does not work.

PR #356, opened 2024-10-18
fix: adjust parseReadalong to preserve the right format

### LINKED ISSUE #347: Cannot download Elan, Praat, SRT or WebVTT from Editor

joanise:
parseFromString does a number of strange and undesirable things:
 - remove the <body> element
 - insert xmlns=
 - change ARPABET= attribute to arpabet= So we need to reverse all this, to match our .readalong DTD format!

Fixes #347

### PR Goal? 

Re-apply #351 changes, but to the reverted main code base.

### Fixes? 

Fixes #347

### Feedback sought? 

check the code, test the fix

### Priority? 

high

### Tests added? 

sadly no

### How to test? 

Open the PR preview, go to the editor, and:
 - try to download Elan or other formats
 - change a word and its alignment and see that it's updated in any download format, i.e., confirm that didn't get broken by this PR
 - you can inspect the parsed .readalong contents by looking at the payload for the eaf call in the Network debugging pane.

### Confidence? 

pretty high

### Version change? 

this and a few other patches will soon trigger patch 1.5.2

### DIFF
--- a/packages/studio-web/src/app/editor/editor.component.ts
+++ b/packages/studio-web/src/app/editor/editor.component.ts
@@ -245,6 +245,17 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
   }
 
   async parseReadalong(text: string): Promise<string | undefined> {
+    // This function is designed to parse three related things:
+    //  - An Offline HTML file with the read-along HTML element with attributes href
+    //    pointing to the .readalong file, and audio pointing to the audio file,
+    //    normally both base64 encoded.
+    //  - The .readalong XML file itself
+    //  - An HTML file with the read-along HTML element and its audio attribute, but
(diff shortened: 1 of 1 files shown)


deltork (approved):
It looks good to me.

roedoejet (approved):
Tested. Works. looks good. thanks @joanise and @deltork !

PR #359, opened 2024-10-18
fix: make the privacy dialogue fully visible on smaller screens

### LINKED ISSUE #352: The privacy policy needs a scroll bar when it doesn't fit on screen, and the analytics opt out buttons becomes inaccessible
On small screens, the policy privacy gets too long to fully display, and there is no scroll bar so you cannot access the Analytics Opt Out / Opt In button at all.

If you try to scroll, the app in the background behind the policy scrolls, but not the policy itself.

joanise:
With a maxHeight of 95vh, we now automatically get a scroll bar if needed, and the Opt out of/in to analytics button is always visible.

Also up the width from 50vw to 60vw, which again is nicer on screens smaller than large monitors.

### PR Goal? 

Make the Opt out of / in to analytics button visible on smaller screens, and add a scroll bar so the whole policy can be read on them.

### Fixes? 

Fixes #352

### Feedback sought? 

sanity checking, more or less just rubber stamping for this one

### Priority? 

medium-high

### Tests added? 

no

### How to test? 

click on Privacy and shrink your viewport to see the Opt in/out button is still usable and the full text can still be accessed.

### Confidence? 

high

### Version change? 

no


roedoejet (approved):
Took the tour, it appears to work well. Thanks!

packages/studio-web/src/app/privacy-dialog.html
```diff
@@ -1,28 +1,33 @@
...
+    >. For a less jargon-y example of the kind of data we see, you can visit
+    this
+    <a target="_blank" href="https://plausible.io/plausible.io"
+      >demonstration site here</a
+    >. In any case, you can opt out entirely by clicking the "Opt Out" button
+    below (you can always opt back in later).
+  </p>
 </div>
 
 <div mat-dialog-actions style="display: flex; justify-content: space-evenly">
```
deltork: Should we add `margin-top: 1em;` to add more space between the text and buttons?
image

deltork (approved):

packages/studio-web/src/app/privacy-dialog.html
joanise: Good idea, I agree.

joanise: Although I'm not sure which element that would belong on...

joanise: And, fixed.

PR #360, opened 2024-10-18
docs: include troubleshooting recommendations in aligner error page

### LINKED ISSUE #335: Add suggestions on the spaceman page: update browser, try Chrome
When a user gets the spaceman error page with Aligner Error, they are told "this may be the result of an incompatible browser" and then "Please contact the devs".

We should modify this with the obvious suggestions first:

> Failed to initialize the aligner. This may be the result of an incompatible browser. Please:
>  - try updating your browser,
>  - try with Chrome.
>
> If the problem persists, please contact the devs...

And maybe if they contact us, we should ask them to specify what browser they tried and which one failed.

There's value in knowing we have incompatibilities, but only if people tell us what we're incompatible with...

joanise:
### PR Goal? 

When SoundSwallower fails to initialize, it's probably an old browser or not Chrome. Give the users a hint to that effect.

### Fixes? 

Fixed #335

### Feedback sought? 

just sanity checking

### Priority? 

low

### Tests added? 

no

### How to test? 

Not sure how to trigger this message in real life, but the way I was testing it locally was going to http://localhost:4204/#/asdf and changing `*ngIf="errorType === 'aligner'"` to `*ngIf="errorType !== 'aligner'"` on line 326 of `error-page-component.html` and the text would get displayed.

### Confidence? 

high

### Version change? 

no


deltork (commented):
I could not test this. However the code changes looks good to me.

deltork (approved):

PR #361, opened 2024-10-18
style: include menu on small devices for toolbar

roedoejet:
### PR Goal? 

Add a menu button instead of the toolbar. Now that I made this though, maybe what we actually want is a "sidenav" element: https://material.angular.io/components/sidenav/overview

### Feedback sought? 

Do you like the way it looks? does it display the languages properly (French/Spanish/English)?

### Priority? 

medium

### Confidence? 

medium-low - maybe we should change this to a sidenav, but I'll leave it here for now so I can continue reviewing the other PRs.


joanise:
Oh, man, that's just so not DRY... :(  But I can still live with it if there's no obvious way to make it DRYer.

Minor presentation detail: in English and Spanish we end up the with the buttons side by side like this 
image
but in French we get
image
which is not nearly as nice. 

So I would add `<br/>` after  `</button>` on lines 17 and 28. I tested, and the result is pleasing, at least to me:
image
image

I'm not convinced sidenav is a better choice, but I'm open to discussion.

deltork:
PR preview is not functional

joanise:
> PR preview is not functional

I just figured it out: we have a hard budget on building the studio-web application, and we've just busted it with this PR. Quite randomly I would say, we've been inching to our budget limit for quite a while now, with 1.98mb and 1.99mb but this PR's 2.01 mb triggering the fatal error.

Fixed now, with a budget of 3mb. Something we ought to try to reduce the size of this, but that's not going to be a priority...

deltork:
> > PR preview is not functional
> 
> I just figured it out: we have a hard budget on building the studio-web application, and we've just busted it with this PR. Quite randomly I would say, we've been inching to our budget limit for quite a while now, with 1.98mb and 1.99mb but this PR's 2.01 mb triggering the fatal error.
> 
> Fixed now, with a budget of 3mb. Something we ought to try to reduce the size of this, but that's not going to be a priority...

Thanks!

packages/studio-web/src/app/app.component.html
```diff
@@ -3,7 +3,37 @@
...
+    <mat-icon>menu</mat-icon>
+  </button>
+  <mat-menu #menu="matMenu">
+    <button
+      i18n="Studio"
+      mat-button
+      [color]="currentURL === '/' ? 'accent' : ''"
+      [routerLink]="''"
+    >
+      Studio
```
deltork: You can add the home icon here. So that the alignment looks good in the mobile view
image

deltork (changes requested):
Looks good! Thanks for taking this on!

packages/studio-web/src/app/app.component.html
joanise: That's an excellent idea. Done, please re-review.

PR #362, opened 2024-10-22
perf: fonts are already base64 text, no need to re-base64 a second time

joanise:
### PR Goal? 

Reduce the size of the Offline HTML file.

### Fixes? 

When we embed the fonts in the Offline HTML, we base64 encode them twice: once from the three `.woff2` files into `fonts.b64.css`, and a second time putting that into the `<link rel="stylesheet" href="...">` line in the `.html` file.

This is wasteful because the `.css` file is already plain text, having had its binary components b64 encoded already.

Instead, this PR proposes to wrap the verbatim contents of the `.css` file in a `<style>` element, which has the same effects.

The results are nice: a trivial readalong goes from 850kb down to 680kb, a saving of roughly 170kb in each RA we download as Offline HTML. 

### Feedback sought? 

testing to make sure this doesn't break anything!

### Priority? 

low

### Tests added? 

no yet, no.

### How to test? 

These are tests I have also done:

Create a readalong or edit an existing one, download it as single file HTML, and see it is smaller but still works correctly, even if you network cable is unplugged.

Upload it back on the currently deployed app, and see you can keep editing, and downloading it from this version or the deployed one, differs in the size of the .html file but not in functionality.

### Confidence? 

Medium-high

### Version change? 

no, because while the .html file is different, it is still fully backwards compatible.

### DIFF
--- a/packages/studio-web/src/app/b64.service.ts
+++ b/packages/studio-web/src/app/b64.service.ts
@@ -35,9 +35,7 @@ export class B64Service {
         ),
       this.http
         .get(this.FONTS_BUNDLE_URL, { responseType: "blob" })
-        .pipe(
-          switchMap((blob: Blob) => this.fileService.readFileAsData$(blob)),
-        ),
+        .pipe(switchMap((blob: Blob) => this.fileService.readFile$(blob))),
     ]);
   }
   utf8_to_b64(str: string) {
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
tested, works. makes a lot of sense! nice catch @joanise !

PR #378, opened 2025-01-07
feat: editor e2e

deltork:
### PR Goal? 
Add tests for the studio editor

### Fixes? 
n/a

### Feedback sought? 
readiness check

### Priority? 
high

### Tests added? 
- https://github.com/ReadAlongs/Studio-Web/blob/main/packages/studio-web/tests/test-cases.md#story-2-main-walk-through-the-editor
- https://github.com/ReadAlongs/Studio-Web/blob/main/packages/studio-web/tests/test-cases.md#take-the-tour-from-the-editor

### How to test? 
* Checkout this branch
* Run studio-cli on your local machine
* Run `npx nx run-many --targets=serve,serve-fr,serve-es --projects=studio-web` in the local branch
* Wait until your browser opens, and you can confirm that studio is up and running
* Run the tests `npx nx e2e studio-web`
* Wait for the report 
#### Test notes
* Firefox tests are currently not 100% stable
* Mobile Chrome test of alignment change is flaky in GitHub ci 
image

### Confidence? 
high

### Version change? 
no, all changes are to support end-to-end testing


packages/studio-web/tests/test-commands.ts
```diff
@@ -131,3 +134,50 @@ export const defaultBeforeEach = async (page: Page, browserName: string) => {
 export const replaceValuesWithZeroes = (text: string): string => {
   return text.replace(/\d/g, "0");
 };
+
+export const disablePlausible = async (page: Page) => {
+  //disable plausible
+  await page.evaluate(
+    async () => await window.localStorage.setItem("plausible_ignore", "true"),
```
joanise, line 141: nice and simple

packages/studio-web/tests/studio-web/check-page-1.spec.ts
```diff
@@ -1,9 +1,10 @@
 import { test, expect } from "@playwright/test";
-import { testText, defaultBeforeEach } from "../test-commands";
+import { testText, disablePlausible } from "../test-commands";
```
joanise, line 2: Nice, thanks for adding this!

joanise:
We should document somewhere that we need to run `npx playwright install --with-deps firefox chromium webkit` before running `npx nx e2e studio-web` for the first time.

joanise:
And also that `npx nx e2e studio-web` tests Studio-Web. The logical place is `### Testing`/`#### Studio-Web` in the top-level README.md.

joanise (approved):
Ready to merge, except for the top-level README.md documenting how to run these tests.

deltork:
> npx playwright install --with-deps firefox chromium webkit

It is part of the husky/post-install process. Should I make it post-checkout instead?

joanise:
The updated readme looks nice, thanks!

joanise:
> > npx playwright install --with-deps firefox chromium webkit
> 
> It is part of the husky/post-install process. Should I make it post-checkout instead?

Post install seems like the right time, as long as it's also documented (as you now did) because if I update an existing sandbox where I did the install before this was added, then I won't have done it.

joanise (approved):
The looks good to me.

The only thing I disagree with is going to 9 shards: we spend about 3 minutes in each shard setting up (install, spin up, etc), and in each shard we do at most 1 minute of work. It just feels like wasted resources. I would go back to 4 shards, maybe even 3, to keep the ratio of setup to work more reasonable. I know this will mean waiting a couple minutes longer for CI to finish, but I don't like all those CI minutes just spent installing.

Is it possible to have multiple threads in the same shard? 
According to https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories the standard runner has 4 CPUs, so maybe it's possible to have just one job doing all the installation just once, and have 4 separate processes in the same job doing all the shards?

joanise:
> so maybe it's possible to have just one job doing all the installation just once, and have 4 separate processes in the same job doing all the shards?

Yes, I believe it is. I just tested and this works in CI:

```
      - name: background and wait
        run: |
          for shard in 1 2 3 4; do
            echo "shard $shard" && sleep 3 &
          done; wait
          echo "all done"
```

Replace the `echo` by the shard command, and then you could run the playwright merge reports command right after in the next step.

joanise:
Oh, wait, playwriting accepts a `--workers 4` option that does the same thing internally, much simpler. Never mind my manual splitting business!

deltork:
> Oh, wait, playwriting accepts a `--workers 4` option that does the same thing internally, much simpler. Never mind my manual splitting business!

I have reduced the shards to 3 and increased the workers to 4.

PR #381, opened 2025-01-22
Dev.ej/license check

joanise:
### PR Goal? 

Add automated license checking to Studio-Web, similar to what we did in g2p and Studio.

### Priority? 

low

### How to test? 

Look at the CI logs, and possibly run these commands interactively:
```
npm install
npx license-checker --production  --onlyAllow 'MIT;0BSD;BSD-2-Clause;BSD-3-Clause;ISC;Apache-2.0' --excludePackages '@nxext/common@18.0.0;unidecode@1.1.0' --summary
npx license-checker  --onlyAllow 'MIT;BSD-2-Clause;ISC;Apache-2.0;0BSD;BSD;Custom: https://nx.dev;Python-2.0;CC-BY-4.0;Apache 2.0;Unlicense;BlueOak-1.0.0;CC0-1.0;CC-BY-3.0' --excludePackages readalong-studio@0.0.0 --summary
```

### Confidence? 

good.

I've reviewed all the licenses I'm listing and they are compatible with MIT to the best of my knowledge.

### Version change? 

no

### DIFF
--- a/.github/workflows/end-to-end-tests.yml
+++ b/.github/workflows/end-to-end-tests.yml
@@ -23,6 +23,15 @@ jobs:
       - name: Install everything
         run: npm install
 
+      - name: Automated license checking
+        env:
+          PROD_LICENSES: "MIT;BSD;0BSD;BSD-2-Clause;BSD-3-Clause;ISC;Apache-2.0"
+          DEV_LICENSES: "Python-2.0;CC0-1.0;CC-BY-3.0;CC-BY-4.0;Apache 2.0;Unlicense;BlueOak-1.0.0"
+          OK_BUT_NOT_AUTODETECTED: "@nxext/common@18.0.0;readalong-studio@0.0.0"
+        run: |
+          npx license-checker --summary --production --onlyAllow "$PROD_LICENSES" --excludePackages "$OK_BUT_NOT_AUTODETECTED"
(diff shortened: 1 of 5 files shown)


joanise:
@deltork You were curious to see how to do license checking in JS, this is what I found.

It's not passing in CI at the moment... It passed on my machine, but it's late so I'll work on fixing it tomorrow.

joanise:
@deltork  OK, license checker is working now. I had missed one license to allow.

The solution I'm picking here means that if we add new dependencies with other allowable licenses, CI will fail and we'll have to review the new license. Since license-checker only lists licenses, without having it's own license compatibility database like the Python version does, this behaviour is desirable: we should inspect each license ourselves and confirm that it's OK.

joanise:
The downside of using npmjs's license-checker package is that it hasn't been updated in 7 years, but since it's only extracting license info and the format for that hasn't changed, it seems to be OK. I'm running it with `npx` without otherwise installing it, so it's in its own isolated environment and using it won't pollute or conflict with our own dependencies.

deltork:
> The downside of using npmjs's license-checker package is that it hasn't been updated in 7 years, but since it's only extracting license info and the format for that hasn't changed, it seems to be OK. I'm running it with `npx` without otherwise installing it, so it's in its own isolated environment and using it won't pollute or conflict with our own dependencies.

Oh, that is interesting!

deltork (approved):
Looks good!

PR #382, opened 2025-01-24
Restore smaller upload file size limits

joanise:
### PR Goal? 

While our old limits of 30KB/60KB were too low to allow realistic Studio-Web use, 250KB is causing crashes. I can crash the server with a 50kb file, even. So restore the max size for text to 40KB, but since the XML is quite bloated, allow 200KB for .readalong inputs.

While I'm here, I also added the same size limit to the text box, so that cut-and-pasting can't be used as a work around for the limit and bring the crashes back.

### Fixes? 

Heroku crash warnings we receive once in a while. Hopefully. 🤞 

### Feedback sought? 

Make sure the code is OK.

### Priority? 

high

### Tests added? 

Normal use is already covered by CI.

Exceeding the limits is not tested, no.

### How to test? 

Spin up locally, or use the PR preview, and submit text file text file > 40kb, or .readalong file > 200kb, and see an error toast.

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "packages": {
     "": {
       "name": "root",
+      "license": "(MIT AND BSD-3-Clause)",
       "workspaces": [
         "packages/*"
       ],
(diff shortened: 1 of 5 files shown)


joanise:
Made this a draft because I forgot to update localization...

joanise:
Localization fixed. Ready to review.

PR #386, opened 2025-01-27
fix: improve the various text-too-big messages

joanise:
### PR Goal? 

Right now, it could be confusing that you are told the file size limit is 40KB, without specifying that it's specifically for text files, but it's different for .readalong files. The message is now tweaked to make it clearer that the limit shown (40KB or 200KB) is for the given file type, Text file or .readalong file.

### Feedback sought? 

general validation and testing.

This is not exercised in CI, so ideally manual testing, even though I'm doing my own as well.

### Priority? 

high, I'd like to deploy this quickly to avoid possible user confusion.

### Tests added? 

max size cases are not in CI yet, but it would be a good idea to add these to e2e testing. @deltork ?

### How to test? 

For each language:
 - select a text file that's >40kb and get an error toast
 - select a .readalong file that's > 200kb and get an error toast
 - paste more than 40kb of text into the text box and click "Go to the next step!" and get an error toast

### Confidence? 

medium-high

### Version change? 

no

### DIFF
--- a/packages/studio-web/src/app/upload/upload.component.ts
+++ b/packages/studio-web/src/app/upload/upload.component.ts
@@ -337,7 +337,8 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
         const inputLength = this.studioService.$textInput.value.length;
         if (inputLength > this.maxTxtSizeKB * 1024) {
           this.toastr.error(
-            $localize`Text too large. Max size: ` +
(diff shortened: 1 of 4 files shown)


deltork:
Can you trigger the size check as you type (or paste)?

deltork (approved):
Looks good!

PR #388, opened 2025-02-13
feat: validate the maximum text size as text is typed of pasted into …

### LINKED ISSUE #387: validate the maximum text size as text is typed of pasted into the text box
With #386 and #382 we now block a request if the text is longer than 40KB, but we only check when the user clicks on Go to the next step!.

Better would be to validate the text size limit as text is entered or pasted.

Related code:
`upload.component.html` line 94:
```
<textarea
                [ngModel]="studioService.$textInput | async"
                (ngModelChange)="studioService.$textInput.next($event)"
``` 

`upload.component.ts` line 89:
```
    this.studioService.$textInput
      .pipe(takeUntil(this.unsubscribe$))
      .subscribe((textString) =>
        this.uploadService.$currentText.next(textString),
      );
```

deltork:
…the text box

### PR Goal? 
1. Added an error message to the text box when a user exceeds the max text size. 
2. Added a test to make sure that validation for text box and text file input being activated

### Fixes? 
Fixes #387 

### Feedback sought? 
validation

### Priority? 
low

### Tests added? 
yes

### How to test? 

Copy the content of `https://github.com/ReadAlongs/Studio-Web/blob/dev.del/issue-387/packages/studio-web/tests/fixtures/ras-text-50kb.txt` into the text box, you should get an error message. 

### Confidence? 
High


joanise:
Tweak request: if you're right at 40kb + 1 byte, and you delete one byte, the toast stays visible, so there's no clear feedback you've fixed things. If you're at 40kb+n bytes, each time you delete a byte a new toast appears, and once you've deleted n bytes, the last toast just stays there, so a user can guess that when the toasts stop flickering, you're OK, but it would be nicer to have the state transition from over to not over generate a green "OK text is now small enough" toast, or something like that.

joanise:
But generally, this is very nice, warning the user as soon as the limit is reached is much nicer than just when they click on next step.

deltork:
> Tweak request: if you're right at 40kb + 1 byte, and you delete one byte, the toast stays visible, so there's no clear feedback you've fixed things. If you're at 40kb+n bytes, each time you delete a byte a new toast appears, and once you've deleted n bytes, the last toast just stays there, so a user can guess that when the toasts stop flickering, you're OK, but it would be nicer to have the state transition from over to not over generate a green "OK text is now small enough" toast, or something like that.

I adjusted the code to dismiss the warning when you go under the limit

joanise:
> I adjusted the code to dismiss the warning when you go under the limit

Did you push the change? I doesn't appear to be fixed in the PR preview.

deltork:
> > I adjusted the code to dismiss the warning when you go under the limit
> 
> Did you push the change? I doesn't appear to be fixed in the PR preview.

It is pushed now

joanise (approved):
Looks and works great, thank you!

PR #389, opened 2025-02-25
feat: use the placeholder to show how to add a page break

joanise:
### PR Goal? 

One step towards addressing the fact that users don't discover how to add pages.

### Feedback sought? 

Does this feel OK? I'm just using the placeholder text, so it disappears when you start typing, but it hints at using newslines as separators.

(We still need to update the tour, though, this is just an extra bit that can already help some.)

### Priority? 

Before ICLDC, if we decide to merge this.

### Tests added? 

n/a

### How to test? 

Load the preview and look at the textbox placeholder text.

### Confidence? 

medium

### Version change? 

no


deltork:
The error generated is because the SRT format output from Studio-cli differs from the sample file we made for the test. One of the commits to CLI has slightly altered the SRT format output.
I just did a fresh download and found that the SRT file has two new lines at the end of the output see attachment (left is ref, right is new download)
image

Solution: replace the `packages/studio-web/tests/fixtures/ref/readalong.srt` with new download or fix cli not to include the extra whitespace

deltork (changes requested):
Looks good!
Maybe fix the CLI SRT output or replace the reference file

joanise:
Interesting, the SRT output with changes to the newlines is the result of changing the version of the webvtt-py package in Studio.

In Studio's unit testing, I added strip() to both strings so tests would not depend on that detail. We'll have to either update the ref file here or do the same, so it's less flaky.

joanise:
I'll try using `trim()`, which appears to be the JavaScript equivalent of Python's `strip()`:
```
  await expect(
    replaceValuesWithZeroes(fileData.replace(/\r/g, "")).trim(),
    "file content should match reference data",
  ).toEqual(replaceValuesWithZeroes(refFileData).trim());
```

joanise:
@deltork Thanks for the analysis on CI. It's fixed with those `.trim()` added on both the output and the reference. This way, the tests will work regardless of the version of Studio and its webvtt-py dependency you have installed.

deltork (approved):

PR #390, opened 2025-02-26
Dev.ej/document how to open

joanise:
### PR Goal? 

When a single-file HTML is opened from Google Drive, you see the raw HTML instead of the file.

We had discussed two solutions, which we both need:
 1- fix how files get opened in Google Drive
 2- have the raw HTML show instructions.

This PR implements 2.

### Feedback sought? 

validation of my help text

### Note

When reviewing, the diff looks big because I also fixed the indentation and used .trim and .replace to de-indent the HTML. Looking at the diff with `git diff -w` is much nicer, highlighting just the new help text.

### Priority? 

before ICLDC

### Tests added? 

was already covered, I'm just changing the specific template strings

### How to test? 

Create an RAS and download the Offline HTML. Upload it to Google Drive, double click on it, and see this:

image

### Confidence? 

High for the implementation, medium for the specific instructions I wrote.

### Version change? 

maybe a patch-level bump

### DIFF
--- a/packages/studio-web/src/app/shared/download/download.service.ts
+++ b/packages/studio-web/src/app/shared/download/download.service.ts
@@ -187,24 +187,51 @@ Please host all assets on your server, include the font and package imports defi
     if (this.b64Service.jsAndFontsBundle$.value !== null) {
       let blob = new Blob(
         [
-          `<!DOCTYPE html>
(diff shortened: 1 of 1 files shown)


deltork (approved):
Works! Thanks for handling this!

PR #391, opened 2025-02-26
feat: custom css interface

deltork:
### PR Goal? 
Allow users to add a custom style sheet to their read along

### Feedback sought? 
the whole works

### Priority? 
High, would be nice for the Eric's presentation

### Tests added? 
Yes

### How to test? 
You can use the test files 
Text: `This is a test.
Sentence.

Paragraph.

Page.`

Audio: `https://github.com/ReadAlongs/Studio-Web/raw/refs/heads/dev.del/feat-studio-add-css/packages/studio-web/tests/fixtures/test-sentence-paragraph-page-56k.mp3`

Stylesheet: `https://raw.githubusercontent.com/ReadAlongs/Studio-Web/refs/heads/dev.del/feat-studio-add-css/packages/studio-web/tests/fixtures/sentence-paragr-cust-css.css`

SingleHTML file: `https://github.com/ReadAlongs/Studio-Web/raw/refs/heads/dev.del/feat-studio-add-css/packages/studio-web/tests/fixtures/sentence-paragr-cust-css.html`

### Confidence? 

High

### Version change? 
minor


joanise:
First impressions: this is great, being able to add fonts and styling. In the Editor, it's good to have it shown below the wavesurfer view. In the Studio, I feel it's a bit too much in your face, for something I expect some users will be excited about, and some users will just want to ignore. Could we have an arrow thing to make showing the Styling editor optional, with the default being close in Studio?

marctessier:
I ran a couple tests and it works for me  and made the interface pink  with the example provided in one example :-) 

To add a comment , can we have the default pre-written in the window ... For example when I start typing in the CSS section, the "example" provided   will erase / I can't copy and paste the "grey" example as the initial template ( If this makes sense what I am saying...)

deltork:
> I ran a couple tests and it works for me and made the interface pink with the example provided in one example :-)
> 
> To add a comment , can we have the default pre-written in the window ... For example when I start typing in the CSS section, the "example" provided will erase / I can't copy and paste the "grey" example as the initial template ( If this makes sense what I am saying...)
Thanks for the review !

If you clicked the help button, you would have seen a sample stylesheet.
@joanise  should we prefill the style?

joanise:
> If you clicked the help button, you would have seen a sample stylesheet.
> @joanise should we prefill the style?

I would say not by default. By default, the style sheet should be empty. But we could have a button that populates it with the template that has all the things you can specify, that would be really helpful.

packages/studio-web/src/app/shared/wc-styling/wc-styling-helper.html
```diff
@@ -0,0 +1,201 @@
...
+          <code>.paragraph</code>
+        </th>
+        <td>The paragraph containing the</td>
+      </tr>
+    </tbody>
+  </table>
+  <h2>Sample Style</h2>
+  <!-- prettier-ignore -->
+  <pre class="css-block">
+    <span class="css-class">:root,
```
joanise: This line should be left-aligned, glued to the margin, so `:root` is shown at the right indent (0-indent) in the help box.

packages/studio-web/src/app/shared/wc-styling/wc-styling-helper.html
```diff
@@ -0,0 +1,201 @@
...
+  --theme-light-system-background: rgb(240, 230, 230) !important;
+  --theme-light-system-foreground: rgb(70, 60, 60) !important;
+  --theme-dark-foreground: rgb(250, 242, 242) !important;
+  --theme-dark-background: rgb(90, 80, 80) !important;
+  --theme-dark-system-background: rgb(80, 70, 70) !important;
+  --theme-dark-system-foreground: rgb(240, 232, 232) !important;
+&#125;
+
+<span class="comment">
+    /**
```
joanise: All the open comment lines with `    /**` should be 0-indented, again so they show correctly in the help.

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.html
```diff
@@ -0,0 +1,254 @@
...
+    </div>
+    <div class="row">
+      <mat-form-field class="col-12 p-0 b-0">
+        <textarea
+          [ngModel]="$styleText | async"
+          (ngModelChange)="$styleText.next($event)"
+          id="styleInput"
+          class="border rounded b-0 p-0 bg-light"
+          matInput
+          placeholder=':root,
```
joanise: Hum, it's nice that the version of this in the help panel has syntax highlighting, but I wish we didn't repeat it twice like this. That's going to be annoying to maintain whenever we make changes. Any chance we can factor the two out, either with automated syntax highlighting there, or with automated stripping of the syntax highlighting here?

And the test suite has a third copy, too. If you had it factored out, the test suite might be able to use it from the same place too...

deltork: I think we could shorten the placeholder, not sure how we can dynamically style the code in the helper

marctessier:
Looks good from my side!

joanise:
As discussed elsewhere, this is good but we would like to do more testing and a more careful review of the code, and maybe polish the interface, so we'll postpone full review merging until after ICLDC.

deltork:
@joanise I need translation help

joanise:
Hmm, when you have a wide screen but a narrow readalong, the ratio is really weird:
image

There are a number of different screen widths where it's not longer behaving like before, I would like to preserve the nice sizing we had settled on, and have this thing just use a small bit of real estate on the side when it's not expanded.

deltork:
> Hmm, when you have a wide screen but a narrow readalong, the ratio is really weird: image
> 
> There are a number of different screen widths where it's not longer behaving like before, I would like to preserve the nice sizing we had settled on, and have this thing just use a small bit of real estate on the side when it's not expanded.

Did you resize the style window? I have modified the code to reset any resizing after you collapse the window

packages/studio-web/src/i18n/messages.json
```diff
@@ -39,6 +39,22 @@
     "6874638171510323302": "WebVTT Subtitles",
     "8762728484338173358": "ReadAlong format conversion failed.",
     "6017683769837067192": "Hmm, we can't connect to the ReadAlongs API. Please try again later.",
+    "6111551964425362195": "Advanced styling",
+    "7226184813051890244": "Advanced Styling",
```
joanise: Please give these two strings to same casing in the source code, so we only have it once in the messages DB.

packages/studio-web/src/app/shared/download/download.service.ts
```diff
@@ -231,7 +240,7 @@ Please host all assets on your server, include the font and package imports defi
                   version="${environment.packageJson.singleFileBundleVersion}"
                   href="data:application/readalong+xml;base64,${rasB64}"
                   audio="${b64Audio}"
-                  image-assets-folder=""
+                  image-assets-folder=""${b64Css}
```
joanise: This confused me being on the same line. Will prettier let you place `${b64Css}` on the next line instead?

packages/studio-web/src/app/shared/download/download.service.ts
```diff
@@ -370,7 +390,7 @@ Please host all assets on your server, include the font and package imports defi
               audio="assets/${basename}.${audioExtension}"
               theme="light"
               language="eng"
-              image-assets-folder="assets/"
+              image-assets-folder="assets/"${pathCss}
```
joanise: ditto

packages/studio-web/src/i18n/messages.json
deltork: fixed

packages/studio-web/tests/test-commands.ts
```diff
@@ -1,5 +1,6 @@
 import { test, expect, Page } from "@playwright/test";
 import process from "process";
+import { __awaiter } from "tslib";
```
joanise: Why this import if you don't use it?

update-translations.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# Run this script to extract new translations from the code and updated the
+# Spanish and French message files with placeholders for missing translations.
```
roedoejet, line 4: Did this file not already exist?

joanise (commented):
This looks good to me now. I think the visual display is good in all sizes, I'm quite happy with the results.
I've looked at the code and it looks fine as far as I can tell, but I'd like Aidan's review before approving, since he'll have a deeper understanding of the implementation that I have.

packages/web-component/README.md
```diff
@@ -180,16 +180,21 @@ and modify only the UI. Use the web inspector of your browser to find the classe
 }
 ```
 
-Here is a list of classes you want to override:
+Here is a list of minimum classes you want to override:
```
roedoejet:
```suggestion
Here is a list of the minimum classes you want to override:
```

packages/web-component/README.md
```diff
@@ -180,16 +180,21 @@ and modify only the UI. Use the web inspector of your browser to find the classe
...
 - .sentence\_\_text.theme--light
+- .sentence\_\_word.theme--dark
+- .sentence\_\_word.theme--dark.reading
+- .sentence\_\_text.theme--dark
 - .sentence\_\_translation
 - .sentence
 - .paragraph
 - .page\_\_container.theme--light (to set page background)
 
+[look at this sample stylesheet for idea](../studio-web/tests/fixtures/sentence-paragr-cust-css.css)
```
roedoejet:
```suggestion
[look at this sample stylesheet to get an idea of what is needed](../studio-web/tests/fixtures/sentence-paragr-cust-css.css)
```

or maybe "for inspiration" or "for an example".

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.sass
```diff
@@ -0,0 +1,70 @@
+mat-dialog-actions.mat-mdc-dialog-actions
+  justify-content: flex-end !important
+
+#styleInput
+  border: 1px solid #222
+  font-family: 'Noto Sans', 'Verdana', 'Arial', 'sans-serif'
```
roedoejet, line 6: Why not use the same font family as defined elsewhere?

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.sass
```diff
@@ -0,0 +1,70 @@
...
+code
+  white-space: pre
+
+.comment
+  color: #696
+
+.css-class
+  color: #ffe66d
+
+.css-block
```
roedoejet, line 29: Did you look into syntax highlighting libraries? I played around a bit with the interface, and while it's nice to have the font be Courier new, I think if we had syntax highlighting, it would further indicate to the user that only CSS should go there (and also the highlighting would help debug issues, especially considering there seem to be a lot of different classes that are required for most use cases.

packages/studio-web/src/app/shared/shared.module.ts
```diff
@@ -4,10 +4,10 @@ import { DownloadComponent } from "./download/download.component";
 import { BrowserModule } from "@angular/platform-browser";
 import { MaterialModule } from "../material.module";
 import { FormsModule } from "@angular/forms";
-
+import { WcStylingComponent } from "./wc-styling/wc-styling.component";
```
roedoejet: If there is only one root-provided service for the component, do we really want to include it in the SharedModule? Since it likely only occurs in the editor, it might be better to define it within `src/editor` and then import it in the app module?

packages/studio-web/src/app/shared/wc-styling/wc-styling-helper.html
```diff
@@ -0,0 +1,214 @@
+<h2 mat-dialog-title i18n="Advanced styling">Advanced styling</h2>
+<mat-dialog-content class="mat-typography">
+  <div class="style-helper">
+    <p>
+      To customize the look and feel of your read-along, you need to upload or
+      write some Cascading Style Sheets in the styling section.
```
roedoejet:
```suggestion
      write some Cascading Style Sheets (CSS) in the styling section.
```

packages/studio-web/src/app/shared/wc-styling/wc-styling-helper.html
```diff
@@ -0,0 +1,214 @@
...
+.color--dark</span> &#123;
+  color: var(--theme-dark-system-foreground) !important;
+&#125;
+
+<span class="css-class">.theme--dark input[type="range"]</span> &#123;
+  accent-color: var(--theme-dark-system-foreground) !important;
+  background-color: var(--theme-dark-system-background) !important;
+&#125;
+
+<span class="css-class">.theme--dark input[type="range"]::-webkit-slider-runnable-track,
```
roedoejet, line 204: This example file is huge! What is this class? Do we need all these example classes? This example feels a bit intimidating.

packages/studio-web/src/app/editor/editor.component.html
```diff
@@ -53,13 +53,32 @@ <h2 i18n="Title for upload box" class="title">
...
+  </div></div-->
+  <div
+    class="d-flex justify-content-center flex-column flex-xl-row flex-xl-nowrap bg-light mt-3 full-height"
+  >
+    <div id="readalongContainer" #readalongContainer></div>
+    <div [class.d-none]="!hasFile" [class.d-lg-flex]="hasFile" class="flex-row">
+      <div
+        class="m-0 p-0 handle me-xl-2 d-none d-xl-block"
+        #handle
+        draggable="true"
```
roedoejet, line 78: The drag has some unexpected behaviour, there's a delay to it, can we get rid of the delay? https://app.screencast.com/hJObrQYDypxli

packages/studio-web/src/app/editor/editor.component.ts
```diff
@@ -51,14 +54,31 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
...
-  ) {}
+    private wcStylingService: WcStylingService,
+  ) {
+    this.wcStylingService.$wcStyleInput.subscribe((css) =>
+      this.updateWCStyle(css),
+    );
+    this.wcStylingService.$wcStyleFonts.subscribe((font) =>
+      this.addWCCustomFont(font),
+    );
+    fromEvent(window, "resize")
```
roedoejet, line 73: oh - I guess this is where the resize lag is coming from. Is this necessary?

packages/studio-web/src/app/editor/editor.component.ts
```diff
@@ -138,8 +192,10 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
           this.readalong,
           this.editorService.slots,
           this.editorService.audioB64Control$.value,
+          this.wcStylingService,
         );
     }
+    this.hasFile = false;
```
roedoejet: Can we make this more descriptive? hasFile is a bit vague. `rasFileIsLoaded` or something specific like that?

packages/studio-web/src/app/editor/editor.component.ts
```diff
@@ -342,6 +409,34 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
...
+    const css = element.getAttribute("css-url");
+    if (css !== null) {
+      if (css.startsWith("data:text/css;base64,")) {
+        this.wcStylingService.$wcStyleInput.next(
+          this.b64Service.b64_to_utf8(css.substring(css.indexOf(",") + 1)),
+        );
+      } else {
+        const reply = await fetch(css);
+        // Did that work? Great!
+        if (reply.ok) {
```
roedoejet, line 441: if this reply isn't ok, it's not handled

packages/studio-web/src/app/editor/editor.component.ts
```diff
@@ -342,6 +409,34 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
...
+          reply.text().then((cssText) => {
+            this.wcStylingService.$wcStyleInput.next(cssText);
+          });
+        }
+      }
+    } else {
```
roedoejet, line 447: minor point, but I like putting smaller conditional blocks ahead of larger ones, so:

```
if (css === null) {
    this.wcStylingService.$wcStyleInput.next("");
} else {
    // all the other stuff
}
```

I think it makes it a bit more readable.

packages/studio-web/src/app/shared/download/download.service.ts
```diff
@@ -371,6 +392,7 @@ Please host all assets on your server, include the font and package imports defi
               theme="light"
               language="eng"
               image-assets-folder="assets/"
+              ${pathCss}
```
roedoejet, line 395: minor point, but I'd rather see `css-url=${pathCss}`

roedoejet: to keep analogous with the rest of the args here

packages/studio-web/src/app/editor/editor.component.html
roedoejet: Also, the divider line z index is too high:

image

And it appears to overflow on the bottom a little too far:

image

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.html
```diff
@@ -0,0 +1,208 @@
...
+  color: var(--theme-dark-foreground) !important;
+}
+
+
+"
+          ></textarea>
+        </mat-form-field>
+      </div>
+      <div class="row">
+        <div class="mt-4 mb-3 ms-5 ms-md-0">
```
roedoejet, line 204: the font upload button disappears if you select "File" for the CSS option

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.html
```diff
@@ -0,0 +1,208 @@
...
+            (ngModelChange)="styleText$.next($event)"
+            id="styleInput"
+            class="border rounded b-0 p-0 bg-light"
+            matInput
+            placeholder="
+Click on the (?) help button above to see a sample stylesheet you can use:
+
+:root,
+#read-along-container {
+  --theme-light-foreground: rgb(80, 70, 70) !important;
```
roedoejet, line 170: Are the only changes that you are really expecting related to the colour and font? If so, maybe we could think about making these into variables instead?

packages/studio-web/src/app/shared/download/download.service.ts
```diff
@@ -232,6 +241,7 @@ Please host all assets on your server, include the font and package imports defi
                   href="data:application/readalong+xml;base64,${rasB64}"
                   audio="${b64Audio}"
                   image-assets-folder=""
+                  ${b64Css}
```
roedoejet, line 244: same here, I'd rather css `css-url=${b64Css}`

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.html
```diff
@@ -0,0 +1,208 @@
...
+        <span class="d-xl-none">
+          <mat-icon>keyboard_arrow_down</mat-icon>
+        </span>
+      </button>
+    </div>
+    <div
+      class="p-0 order-1 order-md-1 helper-text"
+      (click)="collapsed$.next(false)"
+    >
+      <h1 i18n="Advanced styling" class="title pt-2">Advanced styling</h1>
```
roedoejet, line 30: I'm not so sure about the animation here when you collapse the panel. When you expand it, there's no animation, which makes it a little jarring compared to the animation when collapsing the section - maybe we can remove the animation altogether?

roedoejet (changes requested):
Wow — this is a lot of work (and it works!). Nicely done!

I’ve left a few comments — most are very minor — and flagged a couple of smaller UI issues with screenshots that I think we should fix before merging.

One more general thought: I wonder whether it’s worth stepping back for a moment to think about whether building a full-fledged code/CSS editor directly in RAS is the right direction long term. When testing the input area, I found it a bit difficult to edit CSS without syntax highlighting, but adding syntax highlighting might edge us further into scope creep (i.e. developing our own in-browser code editor).

A couple of alternative directions that might be worth considering down the line:

Option A: Rather than having an open CSS input field, we could support uploading CSS files and link users to an external editor (e.g., CodePen) that supports SCSS, syntax highlighting, etc.

Option B: Alternatively, we could narrow the focus to theming (e.g., light/dark foreground/background colours). Users could choose colours via a picker, and we’d generate the CSS under the hood. Not sure if that would fully support your use case, but it might simplify the UX.

To be clear, I’m not suggesting these as replacements for this PR — just possible future directions that could help keep things streamlined while still giving users flexibility. Happy to chat more if helpful!

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.sass
roedoejet: see my general comment: I don't think we need to do this for this PR

packages/studio-web/src/app/shared/download/download.service.ts
deltork: This is done so that the app does not receive `css-url=""` when no style is attached to the web component. In the scenario where there is no custom stylesheet `pathCss=""`, so the tag looks like

 `<readalong ..... image-assets-folder="assets/">`  

instead of

 `<readalong ..... image-assets-folder="assets/" css-url="">`

update-translations.sh
joanise: It sure should have, but no, it did not. Thanks, Del, for adding it!

packages/studio-web/src/app/editor/editor.component.ts
joanise: It fixes a bug I reported earlier, where the app would remember the size you had set it to when you went to xs/s/m/l and the css editor window didn't use the full width. Recalculating when you resize fixes that, at the cost of the lag you see.

deltork:
> Wow — this is a lot of work (and it works!). Nicely done!
> 
> I’ve left a few comments — most are very minor — and flagged a couple of smaller UI issues with screenshots that I think we should fix before merging.
> 
> One more general thought: I wonder whether it’s worth stepping back for a moment to think about whether building a full-fledged code/CSS editor directly in RAS is the right direction long term. When testing the input area, I found it a bit difficult to edit CSS without syntax highlighting, but adding syntax highlighting might edge us further into scope creep (i.e. developing our own in-browser code editor).
> 
The point of the editor is to have a way to preview CSS code you are developing with an IDE. There are very good CSS editors available, so we won't attempt to create one. However, after you have created your custom style sheet, you need a way to "quickly" preview your changes that is component seek to do

I will add a copy and paste button so it is easy to transfer the style between your editor and the box.

> A couple of alternative directions that might be worth considering down the line:
> 
> Option A: Rather than having an open CSS input field, we could support uploading CSS files and link users to an external editor (e.g., CodePen) that supports SCSS, syntax highlighting, etc.
> 

> Option B: Alternatively, we could narrow the focus to theming (e.g., light/dark foreground/background colours). Users could choose colours via a picker, and we’d generate the CSS under the hood. Not sure if that would fully support your use case, but it might simplify the UX.
> 
I am in favour of this option as an add-on (feature for the future, step one is #420, then  #438). This will be a use case for your basic user, and the advanced user will use the pattern I described above. 

> To be clear, I’m not suggesting these as replacements for this PR — just possible future directions that could help keep things streamlined while still giving users flexibility. Happy to chat more if helpful!

Thanks for the thoughtful review. I will incorporate the necessary changes!

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.sass
deltork: This styling is for the help modal only (where sample CSS code is displayed). Designed to make it readable. The actual CSS styling box is simply plain text; it is not an editor, but rather a way to inject your CSS code into your read-along.

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.html
deltork: This is just a sample for the user to spark the imagination

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.html
deltork: Good point! I will remove the animation

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.sass
deltork: I have made an issue for this, see #437

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.html
deltork: It returns when the CSS file is loaded

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.html
deltork: Removed

packages/studio-web/src/app/shared/wc-styling/wc-styling.component.html
roedoejet: Why not keep them both visible at all times? It feels a bit confusing to have it disappear until the CSS file is loaded

roedoejet (approved):
looks good to me, @deltork ! thanks for this

packages/studio-web/src/app/editor/editor.component.ts
deltork: DONE

packages/studio-web/src/app/editor/editor.component.ts
deltork: DONE

packages/studio-web/src/app/editor/editor.component.ts
deltork: DONE

packages/studio-web/src/app/shared/shared.module.ts
deltork: Initially, this module was used in the Demo and Editor, but now that we have restricted it to the Editor, I agree that we should remove it from the SharedModule

deltork: DONE

joanise (approved):
This is ready to merge. No need to rebase again, just create a merge commit as is, you're close enough to the tip.

PR #395, opened 2025-04-07
Dev.ej/streamline bundle

joanise:
### PR Goal? 

This work was triggered by noticing that a .html file someone just sent us was hard to analyse by hand, in terms of what versions of things were used. So I decided to add newlines so that the text part of stuff is still visible when you disable line-wrapping, and add more versioning information to the generated files.

Also, the web-component script does not need to be base64-encoded, that's just bloat: include it directly as is instead.

And in the script tag, add a `name="@readalongs/web-component"` attribute to better document where the script came from.

See also https://github.com/ReadAlongs/Studio/pull/265 for related changes to the CLI.

### Feedback sought? 

General validation.

Question: is there a way to apply my indent in the pipe and switchmap business in b64.service.ts, instead of at download time, so that it could happen in the background earlier?

### Priority? 

normal

### How to test? 

Generate a readalong, download the offline HTML, open it and choose view source to inspect it.

### Confidence? 

good

### Version change? 

no

### DIFF
--- a/packages/studio-web/src/app/b64.service.ts
+++ b/packages/studio-web/src/app/b64.service.ts
@@ -23,7 +23,10 @@ export class B64Service {
     private fileService: FileService,
   ) {
     this.getBundle$().subscribe((bundle) => {
-      this.jsAndFontsBundle$.next(bundle);
+      this.jsAndFontsBundle$.next([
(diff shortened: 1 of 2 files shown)


deltork (approved):
Looks good!

PR #403, opened 2025-05-21
chore(deps): migrate to angular version 19

sergeleger:
### PR Goal? 

Upgrade Angular from version 18 to 19 and the associated Nx dependencies from 19 to 21.

### Fixes? 

Nothing fixed and the migration to Angular 19 required only minor code refactoring.

### Priority? 

Low priority since this work is blocked by a `@nxext/stencil` [issue](https://github.com/nxext/nx-extensions/issues/1180).

### Tests added? 

None.

### How to test? 

All existing tests are passing (with some flakiness). A manual test of the application would be appreciated to ensure I did not miss anything important.

### Confidence? 

Fairly confident - but currently blocked by a `@nxext/stencil` [issue](https://github.com/nxext/nx-extensions/issues/1180)

### Version change? 

No new code or API changes, so a patch at a minimum (1.5.3). Unless, there's a need to reflect a major upgrade of Angular as a minor upgrade (1.6.0).


sergeleger:
@nxext/stencil issue has been resolved, this PR is ready for review.

deltork (approved):
It looks good! However I did get the following warning:
 
```npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated abab@2.0.6: Use your platform's native atob() and btoa() methods instead
npm warn deprecated domexception@4.0.0: Use your platform's native DOMException instead
```

Should we patch the version number of the studio-web ?

sergeleger:
Yes, those warnings have been around this I started working on studio-web. All are from transitive dependencies and are dev dependencies.

joanise (approved):
Wow, that's a lot of dependencies updated!

I agree with Del, we can merge this now.

I also agree we should release a patch. We've fixed quite a few bugs -- thanks Serge!

This would be a patch bump to 1.5.3.

When we add the CSS editor, that will warrant a minor bump to 1.6.0.

joanise:
BTW, `npm audit fix` wants to bump nx to 21.2.0, published just five days ago, but doing so gives me `NX   Unknown feature name 'es6-module'.` when I spin up the studio-web locally.

Maybe it's because I have node 18? Oh, wow, 22 is the LTS, 24 the current, I guess I should bump up to 22.

sergeleger:
We had discussed adding the `engines` parameter to package.json:

```json
{ 
  "engines": {
    "node": ">=20.19.2"
  }
}
```

Do we prefer node 22 instead?

joanise:
Oh, right. Well, with ng 19 or maybe just nx 21.2, we'll need node >=20. I'm in favour of declaring it.

I would declare what we are compatible with, but bump our test and production to 22 since it works cleanly and we won't have to think about the node version for longer if we do.

joanise:
Am I contradicting things I've said before? My apologies if I am, I did not remember having the discussion.

sergeleger:
It was a very brief discussion... we had not specified a node version to support, but as you commented, NX requires node>=20, and I'm comfortable with that. 

Does it make sense here? or in your PR?

joanise:
I guess my PR is touching that kind of stuff, so I can add it there. OK with `"node": ">=20"`? I prefer to declare the lowest we're still compatible with, rather than be aggressive. Otherwise `npm install` will fail unnecessarily. While on the other hand in CI I want to use the latest fully patched version, which is what I get with just `22` or `20`.

joanise:
So I'll do it in my PR, once I can sit down and get to it. My attention is too divided...

PR #404, opened 2025-05-26
fix: audio file re-selection

### LINKED ISSUE #365: Studio: cannot select an audio file, delete it, and re-select it
Steps to reproduce:
 - Launch Studio
 - Click on Audio / File / Choose File
 - Pick a file
 - Click on Delete
 - Click on Choose File and re-select the same file
Result: no file is loaded
Expected result: the same file is re-loaded and usable

Workaround: load a different file, then you can reload the original file.

sergeleger:
### PR Goal? 

Fixes the issue when a user selects an audio file then deletes it; they can now reselect the same audio file.

### Fixes? 

Fixes #365 

### Confidence? 

High.

### DIFF
--- a/packages/studio-web/src/app/upload/upload.component.html
+++ b/packages/studio-web/src/app/upload/upload.component.html
@@ -151,6 +151,7 @@ Page two."
             >Use pre-recorded audio from an MP3 or WAV file.</label
           >
           <input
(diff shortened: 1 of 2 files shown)


joanise (approved):
Thank you, works well!

PR #406, opened 2025-05-28
fix: text file broken in bundle

### LINKED ISSUE #392: The plain text file included in the web bundle is broken in Editor
When you download the bundle from Studio, you get the contents of the text in Step 1 with this code in `download.service.ts`:
```
      // - add plain text file
      if (this.uploadService.$currentText.value !== null) {
        innerFolder?.file(
          `${basename}.txt`,
          this.uploadService.$currentText.value,
        );
      }
```

However, when you download the bundle from Editor, the plain text still comes from this code, fetching whatever text you had in Studio/Step1 before you clicked on Editor, instead of the text that's in the readalong you just uploaded into the Editor.

Short-term hack: we should remove generation of this file when you're in the Editor until this bug is fixed.

Challenge: we don't include the plain text in the single-file HTML, so when we upload to the editor we don't have access to it.

Possible solution: extract the text from the readalong XML structure, reconstructing page, paragraph and sentence boundaries with carefully counted newlines.

sergeleger:
### PR Goal? 

Fixes the broken (or missing) text file when a bundle is generated from the Editor component.

### Fixes? 

Fixes #392 

### Priority? 

Anytime, non-blocking.

### Tests added? 

Yes, added a single web bundle download test for the Editor component.

### How to test? 

From the Editor: 
- upload a previously generated HTML readalong file. 
- download a web bundle
- the generated web bundle should now contain a text file under the `www/` directory with the original user provided text.

### Confidence? 

Fairly confident.

### DIFF
--- a/packages/studio-web/src/app/shared/download/download.service.ts
+++ b/packages/studio-web/src/app/shared/download/download.service.ts
@@ -263,7 +263,7 @@ Please host all assets on your server, include the font and package imports defi
     rasXML: Document,
     slots: ReadAlongSlots,
     readalong: Components.ReadAlong,
-    from = "Studio",
+    from: "Studio" | "Editor" = "Studio",
   ) {
     if (selectedOutputFormat == SupportedOutputs.html) {
       var element = document.createElement("a");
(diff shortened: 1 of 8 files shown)


joanise (approved):
Looks good, and works well, thank you!

Thanks for stabilizing some tests, too, with our own declared selector for download-formats.

PR #408, opened 2025-06-04
fix: standardized export file names across all export formats

### LINKED ISSUE #366: S-Web: improve the filename for the four export formats
In Studio-Web, and presumably in the Editor too, when you choose eaf/srt/vtt/textgrid as the download format and click on the download icon, you get a file called `readalong.<ext>`. We should follow the file naming convention we use for Offline-HTML and Web bundle instead.

sergeleger:
### PR Goal? 

Fixes #366 by standardizing the export filenames across all download types. Additionally, it now uses a timestamp set to the user's local timezone, instead of UTC.

### Priority? 

Small fix and non-blocking, not urgent.

### Tests added? 

Updated existing test to verify new file name format.

### How to test? 

Create a new ReadAlong or upload an existing one in the Editor. Download one of the following formats: eaf/srt/vtt/textgrid. The download file should follow the same naming convention as "Offline HTML" and "Web Bundle" which is:

`{readalong-title}-{timestamp}.{extension}`

The `timestamp` portion should also be in your local timezone (instead of UTC).

### Confidence? 

It is ready to merge.

### DIFF
--- a/packages/studio-web/src/app/shared/download/download.service.ts
+++ b/packages/studio-web/src/app/shared/download/download.service.ts
@@ -250,8 +250,7 @@ Please host all assets on your server, include the font and package imports defi
   }
 
   createRASBasename(title: string) {
-    const timestamp = new Date()
-      .toISOString()
+    const timestamp = dateToLocalISO(new Date())
       .replace(/[^0-9]/g, "")
       .slice(0, -3);
     return (title ? slugify(title, 15) : "readalong") + `-${timestamp}`;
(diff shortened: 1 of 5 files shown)


deltork:
The PR-preview link is broken

joanise:
> The PR-preview link is broken

It works for me.

joanise (approved):
Nice, thank you! Works well.

deltork (approved):
Looks good!

deltork:
> > The PR-preview link is broken
> 
> It works for me.

Yes, working for me now, ah, maybe network issue on my end?

PR #411, opened 2025-06-19
fix: CodeQL XSS security issue

sergeleger:
### PR Goal? 

Introduces a fix to the CodeQL XSS warning.

### Fixes? 

Corrects CodeQL security alert: [DOM text reinterpreted as HTML](https://github.com/ReadAlongs/Studio-Web/security/code-scanning/1)

The mime-type of uploaded page images is now validated. If a user attempts to upload a non-image, the file is silently refused.

Potential memory leak when using `URL.createObjectURL()`. When an image is removed from ReadAlong, `URL.revokeObjectURL()` is called to release the allocated memory. 

### Priority? 

Medium, non-blocking. 

### Tests added? 

None.

### How to test? 

Attempt to upload a non-image file to the read along component. 

### Confidence? 

Medium to high. I'm not certain this will fix the CodeQL warning. I've locally executed codeql on both the main branch and this PR.  My interpretation of that analysis is that the issue has been fixed.

### Version change? 

Maybe included in the upcoming 1.5.3?

### DIFF
--- a/packages/web-component/src/components/read-along-component/read-along.tsx
+++ b/packages/web-component/src/components/read-along-component/read-along.tsx
@@ -1395,18 +1395,24 @@ export class ReadAlongComponent {
     //console.log(JSON.stringify(this.translations));
   }
 
-  async handleFiles(event: any, pageIndex: number) {
-    // const reader = new FileReader()
-    let imageURL = URL.createObjectURL(event);
-    let newImage = {};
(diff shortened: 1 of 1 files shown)


joanise (commented):
Looks good, and works well.

It would be better if we could show some error message when we reject the image file, though.

joanise (approved):
Looks good, with messages for users deferred to #413

Issue #412, opened 2025-06-20
Studio-Web: SASS has deprecated @import statements


sergeleger:
The SASS support for `@import` statements is deprecated:

```
./packages/studio-web/src/styles.sass - Warning: Module Warning (from ./node_modules/sass-loader/dist/cjs.js):
Deprecation Warning on line 3, column 8 of file:///C:/Users/joanise/sandboxes/Studio-Web/packages/studio-web/src/styles.sass:3:8:
Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.
```

PR #416, opened 2025-06-26
fix: corrected Sass @import deprecation warning

sergeleger:
### PR Goal? 

Corrects the Sass `@import` deprecation warning in Studio Web.

### Fixes? 

#412 

### Priority? 

Not urgent, non-blocking and @import's support is still available.

### Tests added? 

None required.

### How to test? 

I've confirmed the deprecation warning is gone. The @import issue was only for ngx-toastr library, so validating that toaster messages are still getting CSS styling would confirm the `@use` statement works correctly.

### Confidence? 

High.


joanise (approved):
looks good

PR #422, opened 2025-07-07
fix: reducing the size of material icon font files

### LINKED ISSUE #350: Filter material fonts for the Studio-Web, Web-C, and the font bundle for Offline HTML
We include the whole of Material fonts in each Offline HTML download, and we load the whole thing each time we load a web-c or the Studio-Web, but we only use about a dozen symbols out of it.

It would be great to filter them.

@dhdaines found these relevant references:
https://the-sustainable.dev/a-guide-to-subsetting-fonts/
https://github.com/zachleat/glyphhanger

sergeleger:
### PR Goal? 

Reduces the size of the fonts asset, containing Material Icon and Material Icon Outlined by requesting only the icons used in the read alongs' component. 

### Fixes? 

Closes #350 

### Feedback sought? 

- Verify that it works as expected. 
- Verify that I did not miss any icons from `web-component` 

### Priority? 

Low priority.

### Tests added? 

Yes, small test (linter) to validate `<MatIcon />` elements do not contain JSX expression.

### How to test? 

- Step 1: Create an "Offline HTML" or "Web Bundle" with the PR preview. 
- Step 2: Create a similar "Offline HTML" or "Web Bundle" with the production site.
- The file generated in Step 1 should be smaller
- Open the file from Step 1 and validate the material icons are present.

### Confidence? 

High, but introduces fetching fonts from Google during the CI pipeline.


packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1164,25 +1165,17 @@ export class ReadAlongComponent {
...
+      "https://fonts.googleapis.com/css2?family=Material+Icons&family=Material+Icons+Outlined&display=swap&icon_names=" +
+      ICON_LIST.join(",");
+    [bcSansFontCssUrl, iconsFontCssUrl]
+      .filter((url) => !document.querySelector(`link[href="${url}"]`))
+      .forEach((url) => {
+        const el = document.createElement("link");
+        el.setAttribute("rel", "stylesheet");
+        el.setAttribute("href", url);
+        document.head.appendChild(el);
+      });
```
joanise, line 1196: nice refactoring!

joanise:
This is marked draft, so I realize I'm looking at it too early, but this is very cool, thanks for figuring this out!

Questions that may already be on your radar -- my apologies if this is all redundant:
 - Can we implement a test in CI that `icons.json` is complete from both studio-web and web-component? 
 - Can we script the generation of `icons.json`? (Not really necessary if CI tests that it's complete, but would be nice.)

joanise:
And I'm perfectly happy with fetching fonts in CI, we already do lots of fetching in there from multiple sources, this is not a significant change.

joanise:
Oh, Studio-Web's download icon is missing
image
as are several others. I'm definitely too early looking at this, sorry. I'll leave you alone until you request a review.

sergeleger:
> Can we script the generation of icons.json? (Not really necessary if CI tests that it's complete, but would be nice.)

This would be ideal solution! and I'm trying... If I succeed, we would no longer need `icons.json` the icon set would be computed by b64Fonts, and I would remove ICON_LIST filtering from read-along.tsx, fixing Studio Web.

Extracting the icon set using Regular expressions... is not appealing, often there's multiline expression or the HTML element contains JS expression (which I would need to parse):

```
 <i class="material-icons" aria-label="Full screen mode">
   {this.fullscreen ? "fullscreen_exit" : "fullscreen"}
 </i>
```

There is no Stencil to AST available, so now I'm looking at JS to AST library...

sergeleger:
Introduced a small functional component to help identify and normalize material icon instances. There was a mix of `<i />` and `<span />` elements used with material icons. 

```html
<MatIcon>delete</MatIcon>
<MatIcon outline>delete</MatIcon>  <!-- uses material-icons-outlined -->
<MatIcon aria-label="Fullscreen window">fullscreen</MatIcon>
```

Read Along no longer filters the icons used; there is no need to add an explicit icon dependency to support Studio Web. Both work as they always have and rely on browser caching to optimize networking.

To produce the font asset bundle, `b64Fonts.js` constructs a regular expression to extract all of the icons used by the read along components. 

**Caveat** There is no support for JSX expressions, and b64Fonts will raise an exception is these are encountered. A common scenario is the following: 

```jsx
<MatIcon>
   {this.fullscreen ? "fullscreen_exit" : "fullscreen"}
</MatIcon>
```

Which can be rewritten as follows:

```jsx
{this.fullscreen ? (
  <MatIcon>fullscreen_exit</MatIcon>
) : (
  <MatIcon>fullscreen</MatIcon>
)}
```

sergeleger:
Since the icon filtering is automated, and `icons.json` does not exist anymore there's no need for a CI check to make sure we've captured all icons.

`b64fonts.js` is executed during the CI pipeline and can fail on JSX expression as described above, I will add a linter / test to capture this error sooner in the development cycle.

joanise (approved):
This is great, thanks!

I like how you lint the usage of MatIcon, it'll save us from confusion in the future when we can't figure out why the heck some icon is visible in dev but not in prod or on a downloaded readalong!

One suggestion: can we add one more check to the linter, namely that `material-icons` should not occurs anywhere in `packages/web-component/src/**/*.ts*` except in `mat-icon.tsx`? That would catch the potential gotcha where one of us wants to add an icon somewhere, and doesn't notice we're using a custom `MatIcon` class to do it.

Besides that, I'm ready to approve.

sergeleger:
Added additional linter rule for uses of either `material-icons` or `material-icons-outlined`.

PR #423, opened 2025-07-07
fix: limit calls to assemble api on sound swallower retries

### LINKED ISSUE #383: When alignment fails with strict or moderate and relaxes, we call assemble again each time
To reproduce, watch the network requests with F12 in Chrome, load a big text file, record a couple seconds of noise, and then click go to next step.

You'll see in the network traffic that assemble is called again from the backend three times, even though it's getting exactly the same output back each time.

The pipeline should be tweaked to retry only the soundswallower alignment step, with updated parameters each time, not the whole pipeline starting from assemble.

This code is in `upload.component.ts`, search for `retry`.

sergeleger:
### PR Goal? 

Eliminates multiple redundant calls to the `/assemble` API. 

### Fixes? 

Closes #383 

### Feedback sought? 

- Confirmation that it works as expected.
- Not very knowledgeable of RxJS, confirmation this is the correct approach. 

### Priority? 

Low, non-blocking.

### Tests added? 

None.

### How to test? 

Follow Eric's instructions from the issue:

> To reproduce, watch the network requests with F12 in Chrome, load a big text file, record a couple seconds of noise, and then click go to next step.
>
> You'll see in the network traffic that assemble is called again from the backend three times, even though it's getting exactly the same output back each time.

Now instead of seeing multiple calls to assemble, there should a single call.

### Confidence? 

High.

### DIFF
--- a/packages/studio-web/src/app/upload/upload.component.ts
+++ b/packages/studio-web/src/app/upload/upload.component.ts
@@ -13,6 +13,8 @@ import {
   map,
   takeUntil,
   throwError,
+  firstValueFrom,
+  take,
 } from "rxjs";
 
 import {
@@ -437,16 +439,19 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
           this.studioService.audioControl$.value as File,
           8000,
         ),
-        ras: this.fileService
-          .readFile$(this.studioService.textControl$.value)
-          .pipe(
-            switchMap((text: string): Observable<ReadAlong> => {
-              body.input = text;
-              this.progressMode = "determinate";
-              this.progressValue = 0;
-              return this.rasService.assembleReadalong$(body);
-            }),
-          ),
+        ras: firstValueFrom(
+          this.fileService
+            .readFile$(this.studioService.textControl$.value)
+            .pipe(
+              switchMap((text: string): Observable<ReadAlong> => {
+                body.input = text;
+                this.progressMode = "determinate";
+                this.progressValue = 0;
+                return this.rasService.assembleReadalong$(body);
+              }),
+              take(1),
+            ),
+        ),
       })
         .pipe(
           switchMap(


joanise:
This reminds me I wanted to improve the messages when alignment fails. They're misleading at the moment. I have a draft of that work somewhere I started when I first created this ticket, I'll have to find it again and actually submit my PR for it.

This PR is good, though. One question, why do we need `take(1)`? We know assemble will always return one result, don't we? From the doc in takeFirstValue they recommend using take, but I don't really understand why, or what could happen if we did not. Anyway, this is me wondering, not a change request: it works correctly as is, though I'd like to understand why.

joanise (approved):

sergeleger:
We'd need Aidan's input,  but my understanding is that `take(1)` releases all subscriptions to the Observable so that it can be destroyed.

joanise:
https://rxjs.dev/api/index/function/firstValueFrom also says it closes the subscription, but my grasp of rxjs is weak enough I can't say I understand if it'll make a difference or not. Is there a way to find out of subscription are leaked or not when debugging such an app?

Anyway, I don't need to understand before we merge, this is just for my education.

sergeleger:
The reason I think this works is the `firstValueFrom` converts the observable to a promise (and `take(1)` then unsubscribes from the observable in effect closing it and releasing its resources). 

A promise can be `await`ed more than once, returning the same resolved value every time:

```
const rasAPIResp = firstValueFrom(...)
const retry1 = await  rasAPIResp;
const retry2 = await rasAPIResp;
const retry3 = await rasAPIResp;
// retry1 === retry2 === retry3
```

`forkJoin` accepts either Observables or Promises, on retry it just does another `await` on the `firstValueFrom` promise which returns the original response.

sergeleger:
in this case `take(1)` might be belt-and-suspenders since we know this observable returns a single value.

joanise:
Thanks for the observable vs promise explanation, that's very helpful, especially since I'm still fuzzy on how they differ.

roedoejet:
I don't think we need the `firstValueFrom` operator here, since we don't handle the result as a promise. take(1) should be sufficient. And yes, you're right @sergeleger - take(N) will just take the first value and then complete the observable.

sergeleger:
Without `firstValuefrom()` the calls to `/assemble` repeats for each SoundSwallower retry (even with `take(1)`). `forkJoin()` does support promises (or PromiseLike), so it does resolve the promise before passing the results to `switchMap`.

PR #430, opened 2025-07-10
fix: added consistent support for xml declarations

### LINKED ISSUE #426: .Readalong export from the web is missing XML declaration
XML declaration is missing from the `.readalong` of the web bundle output, please 
 add the `<?xml version='1.0' encoding='utf-8' ?>` to start of line https://github.com/ReadAlongs/Studio-Web/blob/bf6f13c25e01b1b0d253f8909a3c22233a86d91a/packages/studio-web/src/app/shared/download/download.service.ts#L348

WordPress and other sites might reject the file if this is missing

sergeleger:
### PR Goal? 

The goal of the PR is to generate the XML declaration when creating a `.readalong` file. With sensible data management, we should not be required to manually create XML declarations, they are contained in the `Document` instance, when it is created with `DOMParser().parseFromString(text, "application/xml")`. 

The `/assemble` API did not generate XML declarations, although other tools in read along CLI does. That has been addressed in ReadAlongs/Studio#277. Added in this PR: the front-end code that calls `/assemble` (`ras.service.ts`) also checks for and adds the XML declaration if it is missing. 

The Editor also needs to verify if the XML string loaded from an "Offline HTML" document contains an XML declaration, and adds one if necessary.
 
### Fixes? 

Closes #426 

### Feedback sought? 

Validation the it properly generates the XML declarations.

### Priority? 

High, a client is waiting for the fix. We've decided to do a release after this PR.

### Tests added? 

Not yet, will investigate the existing tests to see if I can add a few additional assertions.

### How to test? 

Unfortunately, the only **easy** verification is with the Web bundle, where we can quickly inspect the `.readalong`.  I've verified that it also works with Offline HTML, but that's a bit harder.

- Create a new read along
- Download the Web bundle
- `.readalong` file should now contain the xml declaration

- Upload an old read along to the Editor
- Download the Web bundle
- `.readalong` file should now contain the xml declaration

### Confidence? 

High, the Editor portion a bit less... but we'll be removing the support for embedded read alongs content.

### Version change? 

Yes? Since we are doing a new release?

### DIFF
--- a/packages/studio-web/src/app/editor/editor.component.ts
+++ b/packages/studio-web/src/app/editor/editor.component.ts
@@ -287,37 +287,63 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
     const parser = new DOMParser();
     const readalong = parser.parseFromString(text, "text/html");
     const element = readalong.querySelector("read-along");
-
     if (element === undefined || element === null) {
       return undefined;
     }
 
-    // Store the element as parsed XML
-    // Create body element, which mysteriously gets removed from the text element.
-    let textNode = element.querySelector("text");
-    if (textNode && !textNode.querySelector("body")) {
-      const body = document.createElement("body");
-      body.id = "t0b0";
-      while (textNode.hasChildNodes()) {
-        // @ts-ignore
-        body.appendChild(textNode.firstChild);
+    // What is the appropriate source for the XML read along document? Either it was
+    // encoded in the element's href attribute, or included as a child element of the
+    // <read-along /> element.
+    //
+    // Prioritize the href implementation since it is more common.
+    const href = element.getAttribute("href");
+    if (href) {
+      const reply = await fetch(href);
+      if (reply.ok) {
+        // FIXME: potential zip-bombing?
+        let xmlString = await reply.text();
+        if (!xmlString.startsWith("<?xml")) {
+          xmlString = `<?xml version='1.0' encoding='utf-8'?>\n` + xmlString;
+        }
+
+        this.editorService.rasControl$.setValue(
+          parser.parseFromString(xmlString, "application/xml"),
+        );
       }
-      textNode.appendChild(body);
+    } else {
+      // Store the element as parsed XML
+      // Create body element, which gets removed from the text element. This occurs
+      // because the document was parsed with text/html mimetype which only allows
+      // a single <body /> element as a child of <html />.
+      let textNode = element.querySelector("text");
+      if (textNode && !textNode.querySelector("body")) {
+        const body = document.createElement("body");
+        body.id = "t0b0";
+        while (textNode.hasChildNodes()) {
+          // @ts-ignore
+          body.appendChild(textNode.firstChild);
+        }
+        textNode.appendChild(body);
+      }
+
+      // Similar issue, the document was parsed with a text/html mimetype, attributes
+      // in HTML are always lowercased.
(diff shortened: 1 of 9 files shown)


packages/studio-web/src/app/editor/editor.component.ts
```diff
@@ -287,37 +287,63 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
...
+        .replace(/arpabet=/g, "ARPABET=") // Our DTD says ARPABET is upper case
+        .replace(/xmlns="[\w\/\:\.]*"/g, ""); // Our DTD does not accept xmlns that the parser inserts
+      if (!xmlString.startsWith("<?xml")) {
+        xmlString = `<?xml version='1.0' encoding='utf-8'?>\n` + xmlString;
       }
-      textNode.appendChild(body);
+
+      this.editorService.rasControl$.setValue(
+        parser.parseFromString(xmlString, "application/xml"),
+      ); // re-parse as XML
```
sergeleger, line 342: These lines will  be removed when we remove support for embedded read along documents.

packages/studio-web/src/app/editor/editor.component.ts
```diff
@@ -287,37 +287,63 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
...
+    // What is the appropriate source for the XML read along document? Either it was
+    // encoded in the element's href attribute, or included as a child element of the
+    // <read-along /> element.
+    //
+    // Prioritize the href implementation since it is more common.
+    const href = element.getAttribute("href");
+    if (href) {
+      const reply = await fetch(href);
+      if (reply.ok) {
+        // FIXME: potential zip-bombing?
```
joanise, line 303: What's the potential zip-bombing issue here?

packages/studio-web/src/app/editor/editor.component.ts
joanise: Either way, you've nicely refactored the code so it's a lot clearer than it used to be.

joanise (commented):
I can't find where that's happening, but in the downloaded bundle the xml declaration and the read-along element are on the same line:
```
<?xml version="1.0" encoding="utf-8"?><read-along version="1.2">
    <meta...
```
but it would be preferable on its own line
```
<?xml version="1.0" encoding="utf-8"?>
<read-along version="1.2">
    <meta...
```

Otherwise this code looks OK and seems to work fine. Thanks for adding relevant unit tests.

packages/studio-web/src/app/editor/editor.component.ts
sergeleger: Not sure, this was a comment left by Aidan which I've kept. 

The previous implementation was recursive, that might have been the issue.

sergeleger:
> I can't find where that's happening, but in the downloaded bundle the xml declaration and the read-along element are on the same line:
> 
> ```
> <?xml version="1.0" encoding="utf-8"?><read-along version="1.2">
>     <meta...
> ```
> 
> but it would be preferable on its own line
> 
> ```
> <?xml version="1.0" encoding="utf-8"?>
> <read-along version="1.2">
>     <meta...
> ```

This is what `XMLSerializer().serializeToString(xml)` generates. I could reformat it in the .readalong file?

joanise:
Looks good to me, thanks for fixing this bug!

joanise (approved):

PR #432, opened 2025-07-14
fix: studio tour shows incorrect user preview 

### LINKED ISSUE #427: Studio-W: warn user that starting a Tour loses all work in Editor
The Studio tour at some point shows the following prompt, warning the user that going to the next step will overwrite the user's data:

Image

The editor tour does not explicitly say it, although the orange button (a warning?) button seems to indicate that was the plan

Image

### LINKED ISSUE #428: Studio-W: during the Studio tour incorrect preview shown to user
If the user had selected the "File" option for either Text or Audio uploads the tour does not show the correct element when providing description:

Image

The description should be associated with the `<textarea />` element and not the file upload element. Clicking "Next" will then properly show the correct description for file uploads.

Steps to reproduce:

- Load Studio Web - Step 1
- Click on "File" for either text file upload or audio file upload
- Start the tour.

sergeleger:
### PR Goal? 

Addresses two minor issues with Studio Web tours. 

In the studio component, the tour text that does not align with the highlighted section of the page.

In the editor component, the tour fails to mention that the current data will be overwritten. 

### Fixes? 

Fixes #428 
Fixes #427 

### Feedback sought? 

Works as expected.

### Priority? 

Low.

### Tests added? 

None.

#### Studio

- Click on both "File" options
- Start the tour.
- When the tour gets to the `<textarea />` section, it should be the element highlighted.
- When the tour gets to the text file upload  section, it should be the element highlighted.
- When the tour gets to the record audio section, it should be the element highlighted.
- When the tour gets to the audio file upload  section, it should be the element highlighted.

Additionally, the tour now reverts these "Write | File" and "Record | File" choices back to what they were prior to the tour. 

#### Editor

- Start a new Tour. 
- At the second step, the tour should now warn the user that their data will be overwritten.

### Confidence? 

High.

### DIFF
--- a/packages/studio-web/src/app/editor/editor.component.ts
+++ b/packages/studio-web/src/app/editor/editor.component.ts
@@ -385,14 +385,6 @@ export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
         .pipe(takeUntil(this.unsubscribe$))
         .subscribe(async (indexFile) => {
           await this.loadRasFile(indexFile);
-          console.log(
-            document
-              .querySelector("#wavesurferContainer")
-              ?.querySelector(".segment-content"),
-            document
-              .querySelector("#readalongContainer")
-              ?.querySelector("read-along"),
-          );
           this.shepherdService.next();
           readalong_add_image_step["attachTo"] = {
             element: document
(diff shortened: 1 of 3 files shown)


joanise (approved):
Looks good, works well.

PR #434, opened 2025-07-16
fix: added explicit utf-8 encoding for PowerShell users

### LINKED ISSUE #415: extract-i18n-lang.ts outputs UCS2 in PowerShell
Running

     npx tsx extract-i18n-lang.ts src/i18n/messages.json src/i18n/messages.fr.json

in PowerShell on Windows, the result is UCS2, which ends up looking like this since we expect utf8:

```
��{�
�
� � �"�l�o�c�a�l�e�"�:� �"�f�r�"�,�
```

Currently, the output is written with 

    console.log(JSON.stringify(languageData, null, 2));

but using

    fs.writeFileSync()

instead would let us explicitly set the encoding to utf8.

This will require adding an argument to the script with the output file name.

sergeleger:
### PR Goal? 

Fixes the `extract-i18n-lang.ts` utility to script, so that it can be used in PowerShell. Using `>` redirect in PowerShell seems generates non-UTF8 files. This PR mitigates this by:

- new `--output` and `--inplace` flags that explicitly writes a utf-8 file without relying on redirections.
- new `--help` flag that describes the tool, with a note for PowerShell users. 
- README.md has been updated to document a first execution of the script with `--help`
- the script attempts to determine if the user's environment is a PowerShell. Logs a warning message if redirection might be in use.
- redirect is still available by omitting `--output` and `--inline`, or specifying `-` as the output file name: `--output -`

### Fixes? 

Fixes #415 

### Feedback sought? 

That it works correctly. Special request for reviewers with access to PowerShell, see How To Test.

### Priority? 

Low, non blocking.

### Tests added? 

No.

### How to test? 

- Test the new `--help` flag and its content:

```bash
npx tsx packages/studio-web/extract-i18n-lang.ts  --help
```

For the rest of the examples, you can create a new translation file by running the following command:
```bash
echo '{"translations": {}}' > test-translation.json
```

- Output translated content to stdout:
```bash
npx tsx packages/studio-web/extract-i18n-lang.ts  packages/studio-web/src/i18n/messages.json test-translation.json
```

- Output translated content to original file with `--inplace`, and report the number of lines in new file (181).
```bash
npx tsx packages/studio-web/extract-i18n-lang.ts --inplace packages/studio-web/src/i18n/messages.json test-translation.json
wc -l test-translation.json
```

- Create a new file with `--output` and compare with source translation file.
```bash
npx tsx packages/studio-web/extract-i18n-lang.ts --output test-translation-updated.json packages/studio-web/src/i18n/messages.json test-translation.json
diff test-translation-updated.json test-translation.json
```

#### Special requests for reviewers with PowerShell

- It's possible that this new implementation now works with PowerShell redirects. The new file should be utf-8 encoded:
```bash
npx tsx packages/studio-web/extract-i18n-lang.ts  packages/studio-web/src/i18n/messages.json test-translation.json > test-translation-updated.json
```
- the previous command should have logged a PowerShell warning. 

#### Cleanup

```bash
rm test-translation.json
rm test-translation-updated.json
```

### Confidence? 

High, except for the PowerShell functionality.

### DIFF
--- a/packages/studio-web/extract-i18n-lang.ts
+++ b/packages/studio-web/extract-i18n-lang.ts
@@ -1,35 +1,106 @@
-// Update the language files with the new keys from the English file
-// Usage:
-//   npx tsx extract-i18n-lang.ts src/i18n/messages.json src/i18n/messages.fr.json > src/i18n/messages.fr-updated.json
-//   npx tsx extract-i18n-lang.ts src/i18n/messages.json src/i18n/messages.es.json > src/i18n/messages.es-updated.json
-// and then manually inspect the results, translate the new keys and copy the
-// updated files back to the original files.
-
 import fs from "fs";
 import process from "process";
+import { parseArgs } from "node:util";
 
-// Get the messages file from the command line
-const englishMessagesFile = process.argv[2];
-const languageMessagesFile = process.argv[3];
+function showHelp() {
+  console.log(`Update the language files with the new keys from the English file.
 
-// Check if filename is provided
-if (!englishMessagesFile || !languageMessagesFile) {
-  console.error(
-    "Please provide the English message file and translated message file as first and second CLI argument",
-  );
-  process.exit(1);
+Then manually inspect the results, translate the new keys and copy the
+updated files back to the original files.
+
+Usage:
+    npx tsx extract-i18n-lang.ts [options] englishFile otherLanguageFile
+
+Options:
+    --output filePath   writes the utf-8 encoded data to the specified output file
+    -i, --inplace       overwrites the translated language file
+    -h, --help              displays this message
+
+Examples:
+    npx tsx extract-i18n-lang.ts src/i18n/messages.json src/i18n/messages.fr.json > src/i18n/messages.fr-updated.json
+    npx tsx extract-i18n-lang.ts src/i18n/messages.json src/i18n/messages.es.json > src/i18n/messages.es-updated.json
+
+    npx tsx extract-i18n-lang.ts --output src/i18n/messages.es-updated.json src/i18n/messages.json src/i18n/messages.es.json
+    npx tsx extract-i18n-lang.ts --inplace src/i18n/messages.json src/i18n/messages.es.json
+
+Note:
+    In a Windows PowerShell environment, please use either the --output or the --inplace flags.
+`);
 }
 
-// Read the English file synchronously
-const englishDataRaw = fs.readFileSync(englishMessagesFile, "utf8");
-const englishData = JSON.parse(englishDataRaw);
-//console.log(englishDataParsed);
+function parseCommandLine() {
+  const { values: flags, positionals: args } = parseArgs({
+    args: process.argv.slice(2),
+    options: {
+      output: {
+        type: "string",
+        default: "-",
+      },
+      inplace: { type: "boolean", short: "i" },
+      help: { type: "boolean", short: "h" },
+    },
+    allowPositionals: true,
+  });
 
-// Read the language file synchronously
-const languageDataRaw = fs.readFileSync(languageMessagesFile, "utf8");
-let languageData = JSON.parse(languageDataRaw);
-//assert(englishData, 'English data is null');
-//assert(languageData, 'Language data is null');
+  // show help message
+  if (flags.help) {
+    showHelp();
+    process.exit(0);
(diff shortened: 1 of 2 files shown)


packages/studio-web/extract-i18n-lang.ts
```diff
@@ -1,35 +1,103 @@
...
+function parseCommandLine() {
+  const { values: flags, positionals: args } = parseArgs({
+    args: process.argv.slice(2),
+    options: {
+      output: {
+        type: "string",
+        default: "-",
+      },
+      inplace: { type: "boolean", short: "i" },
+      help: { type: "boolean" },
```
joanise: I always add `short: "h"` for `--help`, or the equivalent, in every CLI I write in any programming language. Am I old-fashioned?

packages/studio-web/extract-i18n-lang.ts
```diff
@@ -1,35 +1,103 @@
...
+  // specified an output file path.
+  if (flags.output === "-" && flags.inplace) {
+    flags.output = args[1];
+  }
 
-// Get the messages file from the command line
-const englishMessagesFile = process.argv[2];
-const languageMessagesFile = process.argv[3];
+  // verify a PowerShell user is using stdout, provide a warning message.
+  if ("PSModulePath" in process.env && flags.output === "-") {
```
joanise: Nice!

packages/studio-web/extract-i18n-lang.ts
```diff
@@ -1,35 +1,103 @@
...
-const languageDataRaw = fs.readFileSync(languageMessagesFile, "utf8");
-let languageData = JSON.parse(languageDataRaw);
-//assert(englishData, 'English data is null');
-//assert(languageData, 'Language data is null');
+function readJSON(filePath: string): TranslationFile {
+  try {
+    const rawData = fs.readFileSync(filePath, "utf8");
+    return JSON.parse(rawData);
+  } catch {
+    console.log(`error: the file "${filePath}" does not exist`);
```
joanise: Is that really the only exception possible? What if JSON.parse raises a syntax error on the contents?

I would have expected something more like `catch e` and `console.log(`error reading $filePath: `+ str(e))` or some valid TS code that prints the caught exception's message.

packages/studio-web/extract-i18n-lang.ts
```diff
@@ -1,35 +1,103 @@
...
+    [key: string]: string;
+  };
+}
 
-// Read the language file synchronously
-const languageDataRaw = fs.readFileSync(languageMessagesFile, "utf8");
-let languageData = JSON.parse(languageDataRaw);
-//assert(englishData, 'English data is null');
-//assert(languageData, 'Language data is null');
+function readJSON(filePath: string): TranslationFile {
```
joanise, line 84: Super nice to factor this out and catch exceptions.

packages/studio-web/extract-i18n-lang.ts
joanise: For what it's worth, I get the warning when I run in `cmd` and Git Bash too, so the test seems to be too broad. It seems tricky to reliably test for this, though, so I'm fine say I don't care and not change it, but if you have a simple reliable solution, I'd be interested.

joanise (approved):
Nice, the code is much improved. You can tell I didn't really know what I was doing when I wrote it... It was among my early fumblings with TS.

Anyway, "Accept with minor revisions" :) See small suggestions below.

packages/studio-web/extract-i18n-lang.ts
sergeleger: I haven't created short flags in a long time (the last time was probably with `getopts`). I'll add `-h`.

packages/studio-web/extract-i18n-lang.ts
sergeleger: That's unfortunate, PSModulePath is defined globally. There's are no other Env variables to determine cmd vs powershell. 

I'll change the implementation to be clearer about the intention of the conditional by using node's `process.platform === "win32"`. I'll update the warning message to address Windows users instead of powershell users.

packages/studio-web/extract-i18n-lang.ts
sergeleger: Done.

PR #435, opened 2025-07-16
fix: corrected file element that loses the filename of the uploaded file

### LINKED ISSUE #429: Studio-W: file names of uploaded content (text or audio) can disappear on user action
Steps to recreate:

- In Studio Step 1
- Upload an audio
- Then click on "Record" - at this point you can see audio is available, you can play, delete, etc.
- Then click on "File" - the file is no longer available:

A similar issue with Text uploads, but clicking on "Write" does not display anything.

Here's the audio example

Image

sergeleger:
### PR Goal? 

In Step 1 of the Studio: the uploaded filename disappears from the UI when a user toggles between "Record" and "File" (or "Text" and "File). The underlying cause is the use of Angular's `ngIf` expression which destroys the associated DOM elements when they are no longer visible. This PR uses CSS classes to toggle between hidden and visible states.

### Fixes? 

Fixes #429

### Feedback sought? 

Still works as expected.

### Priority? 

Low.

### Tests added? 

None.

### How to test? 

In Studio Step 1
- Upload an audio file
- Then click on "Record" - at this point you can see audio is available, you can play, delete, etc.
- Then click on "File" - the original file name is still visible.

### Confidence? 

High


packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -361,195 +394,220 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
...
+    if (!isValid) {
+      return;
+    }
+
+    if (this.studioService.inputMethod.text === "edit") {
+      const inputText = new Blob([this.studioService.$textInput.value], {
+        type: "text/plain",
+      });
+      this.studioService.textControl$.setValue(inputText);
+    }
```
sergeleger, line 487: I think this block is not required... the `textControl$` gets populated as a blob with the textInput observable (~line 133).

deltork (approved):
This looks good!

PR #440, opened 2025-07-21
Dev.ej/prep v1.5.3

joanise:
### PR Goal? 

We have done lots of improvements to the app, mostly stability and dependency updates, including a lot of bug fixes and migrating to Angular 19.

I think all this warrants an (overdue) patch release, so this PR creates 1.5.3.

With the CSS feature in #391, that would warrant a minor bump to 1.6.0, but I'd like to hag a tag back to the fully patched app before that new feature is in.

### Feedback sought? 

Agreement that this is the right time for a patch.

### Priority? 

high

### Confidence? 

high

### Version change? 

that's what it's all about! :)


deltork (approved):
Yeah this is the right time to publish 1.5.3

PR #442, opened 2025-07-22
ci: make the npm release last since it is not reversible

joanise:
### PR Goal? 

Since GitHub releases can be delete but npmjs published packages cannot, the `npm publish` steps should be the very last ones we run.

This way, when something fails, we can just delete the release on GitHub and push the tag again.

### Feedback sought? 

Sanity checking

### Priority? 

High - before we publish 1.6.0

### How to test? 

Can't test this without creating a fake repo and all, but I'm just interchanging two blocks in the workflow, so I think it's safe.

### Confidence? 

high

### Version change? 

no


sergeleger (approved):
Looks good to me.

roedoejet (approved):
good idea

PR #446, opened 2025-07-23
fix(build): access stylus from github instead of npmjs due to admin error

joanise:
### PR Goal? 

Work around the fact that npmjs banned stylus, purportedly by accident.

See 
https://www.bleepingcomputer.com/news/security/npm-accidentally-removes-stylus-package-breaks-builds-and-pipelines/ https://github.com/stylus/stylus/issues/2938

I'm submitting this PR for info, but the better solution is to wait for them to fix things.

### Priority? 

Del is demo'ing the CSS editor Monday to a community, so if npmjs doesn't fix things soon, we might need to merge this PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28616,8 +28616,7 @@
     },
     "node_modules/stylus": {
       "version": "0.64.0",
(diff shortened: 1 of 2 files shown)


joanise:
Issue fixed on nmpjs. Closing.

PR #450, opened 2025-08-05
fix: corrected all deprecated API calls including ngIf and ngFor

### LINKED ISSUE #418: Angular 20 deprecates ngIf attribute
As reported by @deltork, `ngIf` is scheduled to be removed in Angular v22. The frequency of major releases are every 6 months, v22 would tentatively land in May 2026.

sergeleger:
### PR Goal? 

Replaces all deprecated API calls, including `nfIg` and `ngFor`.  Additionally, this commit was created in the larger context of early state management work, so it includes class member visibility declarations (private / protected / public).

### Fixes? 

Closes #418 

### Feedback sought? 

Application still works as expected.

### Priority? 

Medium to high, since the visibility declarations touched all classes it would be nice to include this work sooner rather than later.

### Tests added? 

No, but many deprecated Angular APIs were used in tests - those have been updated.

### Confidence? 

High, I've been using this commit for a while now.

### DIFF
--- a/packages/studio-web/src/app/demo/demo.component.html
+++ b/packages/studio-web/src/app/demo/demo.component.html
@@ -1,62 +1,69 @@
 <section>
-  <div class="container" *ngIf="studioService.render$ | async">
-    <div class="row">
-      <div class="col col-md-6 col-lg-9">
-        <h2
-          id="readalong-congratulations"
-          i18n="Congratulations message"
-          class="subtitle"
-        >
-          Congratulations! Here's your ReadAlong!
-        </h2>
+  @if (studioService.render$ | async) {
+    <div class="container">
+      <div class="row">
+        <div class="col col-md-6 col-lg-9">
+          <h2
+            id="readalong-congratulations"
+            i18n="Congratulations message"
+            class="subtitle"
+          >
+            Congratulations! Here's your ReadAlong!
+          </h2>
+        </div>
+        <div class="col col-md-6 col-lg-3">
+          <ras-shared-download
+            (downloadButtonClicked)="download($event)"
+          ></ras-shared-download>
+        </div>
       </div>
-      <div class="col col-md-6 col-lg-3">
-        <ras-shared-download
-          (downloadButtonClicked)="download($event)"
-        ></ras-shared-download>
-      </div>
-    </div>
 
-    <div class="row"></div>
+      <div class="row"></div>
 
-    <div class="row" *ngIf="studioService.b64Inputs$ | async as b64Inputs">
-      <div *ngIf="b64Inputs[0] && b64Inputs[1] && rasAsDataURL()">
-        <read-along
-          id="readalong"
-          [language]="language"
-          #readalong
-          *ngIf="studioService.render$ | async"
-          mode="EDIT"
-          href="{{ rasAsDataURL() }}"
-          audio="{{ b64Inputs[0] }}"
-          class="hydrated"
-        >
-          <input
-            #title
-            [(ngModel)]="studioService.slots.title"
-            [ngStyle]="{
-              'width.ch': studioService.slots.title.length,
-              'min-width.ch': 20,
-            }"
-            style="border: none"
-            placeholder="Enter your title here"
-            slot="read-along-header"
-            data-test-id="ra-header"
-          />
-          <input
-            #subtitle
-            [(ngModel)]="studioService.slots.subtitle"
-            [ngStyle]="{
-              'width.ch': studioService.slots.subtitle.length,
-              'min-width.ch': 20,
-            }"
-            style="border: none"
-            placeholder="Enter your subtitle here"
-            slot="read-along-subheader"
-            data-test-id="ra-subheader"
-          />
-        </read-along>
-      </div>
+      @if (studioService.b64Inputs$ | async; as b64Inputs) {
+        <div class="row">
+          @if (b64Inputs[0] && b64Inputs[1] && rasAsDataURL()) {
+            <div>
+              @if (studioService.render$ | async) {
+                <read-along
+                  id="readalong"
+                  [language]="language"
+                  #readalong
+                  mode="EDIT"
+                  href="{{ rasAsDataURL() }}"
+                  audio="{{ b64Inputs[0] }}"
+                  class="hydrated"
+                >
+                  <input
+                    #title
+                    [(ngModel)]="studioService.slots.title"
+                    [ngStyle]="{
+                      'width.ch': studioService.slots.title.length,
+                      'min-width.ch': 20,
+                    }"
+                    style="border: none"
+                    placeholder="Enter your title here"
+                    slot="read-along-header"
+                    data-test-id="ra-header"
+                  />
+                  <input
+                    #subtitle
+                    [(ngModel)]="studioService.slots.subtitle"
+                    [ngStyle]="{
+                      'width.ch': studioService.slots.subtitle.length,
+                      'min-width.ch': 20,
+                    }"
+                    style="border: none"
+                    placeholder="Enter your subtitle here"
+                    slot="read-along-subheader"
+                    data-test-id="ra-subheader"
+                  />
+                </read-along>
+              }
+            </div>
+          }
+        </div>
+      }
     </div>
-  </div>
+  }
 </section>
--- a/packages/studio-web/src/app/demo/demo.component.ts
+++ b/packages/studio-web/src/app/demo/demo.component.ts
@@ -6,6 +6,13 @@ import { StudioService } from "../studio/studio.service";
 import { DownloadService } from "../shared/download/download.service";
 import { SupportedOutputs } from "../ras.service";
 import { ToastrService } from "ngx-toastr";
+import { type InterfaceLanguage } from "@readalongs/web-component";
+
+const localizationToRASLanguage: Record<string, InterfaceLanguage> = {
+  en: "eng",
+  fr: "fra",
+  es: "spa",
+};
 
 @Component({
   selector: "app-demo",
@@ -16,7 +23,7 @@ import { ToastrService } from "ngx-toastr";
 export class DemoComponent implements OnDestroy {
   @ViewChild("readalong") private readalong!: Components.ReadAlong;
 
-  protected language: "eng" | "fra" | "spa" = "eng";
+  protected language: InterfaceLanguage = "eng";
   protected b64Service = inject(B64Service);
   protected rasAsDataURL = signal<string>("");
   public studioService = inject(StudioService);
@@ -24,11 +31,9 @@ export class DemoComponent implements OnDestroy {
   private toastr = inject(ToastrService);
 
   constructor() {
-    // If we do more languages, this should be a lookup table
-    if ($localize.locale == "fr") {
-      this.language = "fra";
-    } else if ($localize.locale == "es") {
-      this.language = "spa";
+    const curLocale = $localize.locale ?? "en";
+    if (curLocale in localizationToRASLanguage) {
+      this.language = localizationToRASLanguage[curLocale];
     }
 
     this.studioService.b64Inputs$.subscribe(async (b64Input) => {
--- a/packages/studio-web/src/app/demo/demo.component.ts
+++ b/packages/studio-web/src/app/demo/demo.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnDestroy, OnInit, signal, ViewChild } from "@angular/core";
+import { Component, inject, OnDestroy, ViewChild, signal } from "@angular/core";
 import { Components } from "@readalongs/web-component/loader";
 
 import { B64Service } from "../b64.service";
@@ -13,16 +13,17 @@ import { ToastrService } from "ngx-toastr";
   styleUrls: ["./demo.component.sass"],
   standalone: false,
 })
-export class DemoComponent implements OnDestroy, OnInit {
-  @ViewChild("readalong") readalong!: Components.ReadAlong;
-  language: "eng" | "fra" | "spa" = "eng";
+export class DemoComponent implements OnDestroy {
+  @ViewChild("readalong") private readalong!: Components.ReadAlong;
+
+  protected language: "eng" | "fra" | "spa" = "eng";
+  protected b64Service = inject(B64Service);
   protected rasAsDataURL = signal<string>("");
-  constructor(
-    public b64Service: B64Service,
-    public studioService: StudioService,
-    private downloadService: DownloadService,
-    private toastr: ToastrService,
-  ) {
+  public studioService = inject(StudioService);
+  private downloadService = inject(DownloadService);
+  private toastr = inject(ToastrService);
+
+  constructor() {
     // If we do more languages, this should be a lookup table
     if ($localize.locale == "fr") {
       this.language = "fra";
@@ -37,9 +38,6 @@ export class DemoComponent implements OnDestroy, OnInit {
     });
   }
 
-  ngOnInit(): void {}
-
-  ngAfterViewInit(): void {}
   download(download_type: SupportedOutputs) {
     if (
       this.studioService.b64Inputs$.value &&
--- a/packages/studio-web/src/app/editor/editor.component.ts
+++ b/packages/studio-web/src/app/editor/editor.component.ts
@@ -8,7 +8,6 @@ import {
   ElementRef,
   inject,
   OnDestroy,
-  OnInit,
   ViewChild,
 } from "@angular/core";
 import SegmentsPlugin from "./segments";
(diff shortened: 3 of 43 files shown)


dhdaines:
Oh! nice work!  I haven't been keeping up with Angular but the new template conditional syntax seems more concise, I hope it is less ambiguous than JSX/TSX.

packages/studio-web/src/app/demo/demo.component.html
```diff
@@ -1,62 +1,71 @@
...
+            <div>
+              @if (studioService.render$ | async) {
+                <read-along
+                  id="readalong"
+                  [language]="language"
+                  #readalong
+                  mode="EDIT"
+                  href="data:application/readalong+xml;base64,{{
+                    b64Service.xmlToB64(b64Inputs[1])
+                  }}"
```
joanise: Changing `href="{{ rasAsDataURL() }}"` to this is not just refactoring, it should probably be in a different PR, unless it was introduced here by accident.

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -7,39 +7,32 @@ import { DownloadService } from "../shared/download/download.service";
...
-  }
+  protected language: rasLanguages = "eng";
+  protected b64Service = inject(B64Service);
+  public studioService = inject(StudioService);
+  private downloadService = inject(DownloadService);
+  private toastr = inject(ToastrService);
 
-  ngOnInit(): void {}
+  constructor() {
+    this.language = localizationToRASLanguage[$localize.locale ?? "en"];
```
joanise: Is this robust to `locallize.locale` being an invalid value? The previous code's if/else would leave it  as the default eng if invalid, we should preserve that behaviour unless there's a mechanism blocking invalid values, but I don't know if there is one.

packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -7,39 +7,32 @@ import { DownloadService } from "../shared/download/download.service";
...
-    } else if ($localize.locale == "es") {
-      this.language = "spa";
-    }
+export class DemoComponent implements OnDestroy {
+  @ViewChild("readalong") private readalong!: Components.ReadAlong;
 
-    this.studioService.b64Inputs$.subscribe(async (b64Input) => {
-      if (b64Input[1]) {
-        this.rasAsDataURL.set(await this.b64Service.rasToDataURL(b64Input[1]));
-      }
```
joanise: I'm not sure why we no longer need to set `this.rasAsDataURL` here, can you clarify?

packages/studio-web/src/app/editor/editor.component.ts
```diff
@@ -45,35 +44,34 @@ import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
...
   @ViewChild("handle") handleElement!: ElementRef;
   @ViewChild("styleWindow") styleElement!: WcStylingComponent;
 
   @ViewChild("rasFileUpload")
   private rasFileUpload: ElementRef<HTMLInputElement>;
-  readalong: Components.ReadAlong;
+  private readalong: Components.ReadAlong;
 
-  language: "eng" | "fra" | "spa" = "eng";
+  private language: "eng" | "fra" | "spa" = "eng";
```
joanise: You define a type for this in `demo.component.ts`, which seems like a good idea, but is it feasible to factor that definition out to one place instead of repeating it?

joanise: The same also occurs in `index.d.ts` but maybe that's a generated file?

Anyway, I would either undo the type definition in demo component, going back to hard-coding the list, or else properly refactor this so it's defined in just one place. I don't like the partial refactoring where we name the type in one of three places.

packages/studio-web/src/app/file.service.ts
```diff
@@ -9,21 +8,19 @@ import {
...
-  constructor(
-    private http: HttpClient,
-    private toastr: ToastrService,
-  ) {}
+  private http = inject(HttpClient);
+  private toastr = inject(ToastrService);
 
   loadAudioBufferFromFile$(
-    file: File,
+    file: File | Blob,
```
joanise, line 24: What does this fix or do?

packages/studio-web/src/app/file.service.ts
```diff
@@ -65,7 +62,7 @@ export class FileService {
     return new Observable((obs: any) => {
       reader.onerror = (err) => obs.error(err);
       reader.onabort = (err) => obs.error(err);
-      reader.onload = () => obs.next(reader.result);
+      reader.onload = () => obs.next(reader.result as string);
```
joanise, line 66: ditto?

packages/studio-web/src/app/file.service.ts
```diff
@@ -91,7 +88,7 @@ export class FileService {
...
```
joanise, line 92: ditto?

packages/studio-web/src/app/material.module.ts
```diff
@@ -12,6 +12,7 @@ import { MatSlideToggleModule } from "@angular/material/slide-toggle";
 import { MatStepperModule } from "@angular/material/stepper";
 import { MatMenuModule } from "@angular/material/menu";
 import { MatDividerModule } from "@angular/material/divider";
+import { MatToolbarModule } from "@angular/material/toolbar";
```
joanise, line 15: This PR really combines too many things, I can't tell what this change is for.

packages/studio-web/src/app/microphone.service.ts
```diff
@@ -10,12 +12,12 @@ export class MicrophoneService {
...
 
   async startRecording() {
-    if (this.recorder !== null && this.recorder.state == "paused") {
+    if (this.recorder !== null && this.recorder.state === "paused") {
       this.resume();
       return;
     }
     this.stream = await navigator.mediaDevices.getUserMedia({ audio: true });
-    this.recorder = new MediaRecorder(this.stream);
+    this.recorder = new MediaRecorder(this.stream, { mimeType: "audio/mpeg" });
```
joanise, line 18: Is this some bug fix?

packages/studio-web/src/app/microphone.service.ts
```diff
@@ -1,5 +1,7 @@
 import { EventEmitter, Injectable } from "@angular/core";
 
+export const ERR_NO_RECORDING = "Recorder didn't hear anything";
```
joanise: Seems pointless to create a constant that's only used in one place.

packages/studio-web/src/app/ras.service.spec.ts
```diff
@@ -34,7 +36,7 @@ describe("RasService", () => {
     });
     const req = httpTestingController.expectOne({
       method: "GET",
-      url: service.baseURL + "/langs",
+      url: `${environment.apiBaseURL}/langs`,
```
joanise: getting this from the service we're testing made more sense to me

packages/studio-web/src/app/soundswallower.service.ts
```diff
@@ -79,7 +79,7 @@ export class SoundswallowerService {
     return new Observable((subscriber) => {
       // Do synchronous (and hopefully fast) initialization
       const decoder = new soundswallower.Decoder({
-        loglevel: "INFO",
+        loglevel: "ERROR", // "INFO"
```
joanise: good idea

packages/studio-web/src/app/studio/studio.component.spec.ts
```diff
@@ -40,10 +44,4 @@ describe("StudioComponent", () => {
     const app = fixture.componentInstance;
     expect(app).toBeTruthy();
   });
-
-  it(`should have as title 'readalong-studio'`, () => {
-    const fixture = TestBed.createComponent(StudioComponent);
-    const app = fixture.componentInstance;
-    expect(app.title).toEqual("readalong-studio");
-  });
```
joanise, line 48: Is this test no longer relevant?

joanise (commented):
Wow, this is a very difficult PR to review. I have read everything carefully, I understand and approve the ngif/ngfor changes, the deprecation fixes make sense, I think I agree with the visibility changes, but I'm not sure I understand all the other changes well enough to approve wholesale. Some of it was simple refactoring, some of it might have been bug fixes?

I apologize for this mitigated answer, a lot of useful work has gone into this, and I'd like to reap the benefit from it all, but I'm not confident enough in all of it to just say go.

packages/studio-web/src/app/demo/demo.component.html
sergeleger: Yeah, that code is not correct, the original `href="{{ rasAsDataURL() }}"` is. `demo.component.html` is one of those files that git could not merge correctly (i.e.: the HTML document semantics were not respected).

packages/studio-web/src/app/demo/demo.component.ts
sergeleger: Agreed, will add an additional conditional to default to "eng" should `$localize.locale` be invalid.

packages/studio-web/src/app/file.service.ts
sergeleger: Mostly a documentation change, based on the usage in the function the actual fix would have been `file: Blob` (since File inherits from Blob ). The function does not access any of File's instances properties (like name, date last modified, etc.)

In this case using `File | Blob` is superfluous, but I like it for documentation purposes.

packages/studio-web/src/app/file.service.ts
sergeleger: The `reader.result` type is dependant on the on which reader method is used. `Reader.readAsArrayBuffer()` returns a Blob, while `Reader.readAsText()` returns a string. In this case we can explicitly declare it as a string which TypeScript should be able to use at the call site.

~I'm worried about the `any` in `new Observable((obs: any)` will investigate to see if this is not another rebase gone awry.~

packages/studio-web/src/app/file.service.ts
sergeleger: `readAsDataURL()` also returns a string.

packages/studio-web/src/app/material.module.ts
sergeleger: This fixes (I think) a compilation warning that occurs when we run studio-web's unit tests.

packages/studio-web/src/app/microphone.service.ts
sergeleger: You are correct, in a previous PR `ERR_NO_RECORDING` was imported by upload.component. That code has since changed but we still have this export. Will remove.

packages/studio-web/src/app/microphone.service.ts
sergeleger: Not a bug (yet) just a clean up I did while refactoring state. 

This explicitly tells `MediaRecorder` we'd like "audio/mpeg", the previous version assumed at line 77 (I'll add a comment below) that we had received audio/mpeg.

packages/studio-web/src/app/microphone.service.ts
```diff
@@ -62,11 +68,13 @@ export class MicrophoneService {
...
   private addListeners() {
     if (this.recorder === null) throw "Recorder was not created";
+
+    const mimeType = this.recorder.mimeType;
     this.recorder.addEventListener("dataavailable", (event: BlobEvent) => {
       this.chunks.push(event.data);
     });
     this.recorder.addEventListener("stop", (event: Event) => {
-      const blob = new Blob(this.chunks, { type: "audio/mpeg" });
+      const blob = new Blob(this.chunks, { type: mimeType });
```
sergeleger, line 71: Now we can use the recorder's mime type (@ 72) to create our blob object (@ 77)

packages/studio-web/src/app/ras.service.spec.ts
sergeleger: Ok, I'll update the code. The reason was that `RasService.baseUrl` is now a private field (it really does not need to be public and available to other code) except for testing purposes.

packages/studio-web/src/app/studio/studio.component.spec.ts
sergeleger: Good question, I forget why I removed it... will investigate.

sergeleger (commented):
I'll investigate more tomorrow. At this point I may just `git reset main` to try to clear some of the changes getting picked up.

packages/studio-web/src/app/editor/editor.component.ts
sergeleger: The `language` variable is not used I've removed it. Will still formalize the type in RasService for now.

sergeleger: `@readalongs/web-component` already defines this type (`InterfaceLanguage`) will import it instead of redefining it.

packages/studio-web/src/app/ras.service.spec.ts
joanise: Ah, I wondered if it might be something like that. I guess it's fine either way, then.

packages/studio-web/src/app/studio/studio.component.spec.ts
sergeleger: Confirmed no longer relevant (imo): `StudioComponent.title` is private and is never used (other than the initialization code). So I have removed `title` from the component. There's already a test to verify the component can be initialized.

joanise:
Great, thank you for splitting this into a bunch of commits.

Checkboxes for my use for reviewing commit by commit:
 * [x] fix: replaced deprecated ngIf and ngFor directives
 * [x] fix: added a language lookup table to convert two-letter codes to three-letter codes
 * [x] fix: lowered the logging verbosity of the sound swallower service
 * [x]  fix: added explicit audio mime type to MediaRecorder
 * [x]  fix: added visibility modifiers to all class properties
 * [x]  fix: added type information
 * [x]  fix: corrected unknown components error while running studio web's unit tests
 * [x]  fix: replaced deprecated Angular testing modules

joanise:
Wow, I just discovered an arcane bug on this PR, but at least it's easy to reproduce:
 - fresh launch of the PR preview
 - type a word in text
 - record a second
 - click Go to the next step
 - click Editor
 - click Studio
 - see `Error: the RAS file could not be parsed`
image

More generically, it looks like you get this error any time you go from Studio step 2 to Editor and back to Studio.

Clicking on Step 1 and then again on Step 2 fixes it: you can see the read along you previously had in step 2 before going to the Editor.

joanise:
It took me a while to figure out what `fix: added a language lookup table to convert two-letter codes to three-letter code` really does: if you have previously saved a language preference using the web component's gear menu, that overrides what's happening here and it makes this code effectively a no-op. But I cleared my site data for our sites and now I can see what it does. @deltork it's a bit of a gotcha that saving your viewing preferences affects the readalongs your create and download later. Is that a bug or desirable?

packages/studio-web/src/app/shared/download/download.service.ts
```diff
@@ -30,8 +30,13 @@ interface Image {
 })
 export class DownloadService {
   private destroyRef$ = inject(DestroyRef);
-  xmlSerializer = new XMLSerializer();
-  readmeFile = new Blob(
+  private xmlSerializer = new XMLSerializer();
```
joanise, line 33: That's not a change you introduced, but I don't understand why we pre-initialize this object when we could use `XMLSerializer()` where we use `this.xmlSerializer`. Especially since we use this pre-initialized instance in only one of the two places where we need it, and create a new one in the other. I would ditch this variable... (but I realize this comment is potential scope creep, don't take it as a change request for this PR, but as technical debt we might consider cleaning)

sergeleger: Agreed, I've seen this example a few times during the summer months. Never did change it since I wasn't sure it was a design decision.

sergeleger:
> Wow, I just discovered an arcane bug on this PR, but at least it's easy to reproduce:
> 
> * fresh launch of the PR preview
> * type a word in text
> * record a second
> * click Go to the next step
> * click Editor
> * click Studio
> * see `Error: the RAS file could not be parsed`
> ...

Confirmed bug was introduced in a previous commit. Will create an issue and fix it in a separate PR.

sergeleger:
> It took me a while to figure out what `fix: added a language lookup table to convert two-letter codes to three-letter code` really does: if you have previously saved a language preference using the web component's gear menu, that overrides what's happening here and it makes this code effectively a no-op. But I cleared my site data for our sites and now I can see what it does. @deltork it's a bit of a gotcha that saving your viewing preferences affects the readalongs your create and download later. Is that a bug or desirable?

At a minimum, this sets a sane default language value for the read along component.

joanise (approved):
This looks good to me now. Thanks for breaking it up, it still took quite a while to review but by reviewing it commit-by-commit I was able to understand what's happening.

Approved with the assumption that you'll fix the conflict with #456 upon merging or rebasing.

Thanks for all this work, it's great to address the technical debt of the code base like this.

PR #451, opened 2025-08-05
refactor: updated B64Service and FileService to prefer FileReader API

sergeleger:
### PR Goal? 

Replaces custom base64 encoding implementation with `FileReader.readAsDataURL()` [MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsDataURL), which creates valid `data:mimetype;base64,` data URLs.  

The web component and font bundles are only fetched when required (by initiating a download). The results are cached for subsequent requests.
 
### Fixes? 

n/a

### Feedback sought? 

Discussion on viability of the proposed solution. We could also officially deprecate and remove previous base64 functions: `utf8_to_b64()`,  `b64_to_utf8()` and `xmlToB64`. 

There's also an opportunity to migrate the remaining data URL methods from `B64Service` to `FileService`, isolating the `getBundle` methods to create a new `BundleService`.

### Priority? 

Low.

### Tests added? 

Yes.

### How to test? 

- The demo component still works
- Offline HTML is still functional
- Web bundle is also functional.

### Confidence? 

Medium.

### DIFF
--- a/packages/studio-web/src/app/file.service.spec.ts
+++ b/packages/studio-web/src/app/file.service.spec.ts
@@ -25,4 +25,33 @@ describe("FileService", () => {
   it("should be created", () => {
     expect(service).toBeTruthy();
   });
+
+  it("should read a text file from a promise", async () => {
+    const got = await service.readFile("this is a test");
+    expect(got).toBe("this is a test");
+  });
+
+  it("should create a data URL from a promise", async () => {
+    const got = await service.readFileAsDataURL("this is a test");
+    expect(got.startsWith("data:text/plain;base64")).toBeTrue();
+
+    // reverse the readFileAsDataURL operation.
+    const reversed = await fetch(got).then((resp) => resp.text());
+    expect(reversed).toBe("this is a test");
+  });
+
+  it("should turn utf8 to b64 and back", async () => {
+    const testUTF8 = `󳬏ۓ脶򍫐䏷򻱚1󔊣򧰗J鷍ˑ⨘󝳗ʳꟴ󆋔=є򻥼Ӳ򦿴¦槩7}摠꾀𴮣۝م𬷊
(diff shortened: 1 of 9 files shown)


packages/studio-web/src/app/file.service.ts
```diff
@@ -42,26 +51,57 @@ export class FileService {
...
+  // A promise based implementation of readFile$.
+  readFile(
+    blob: Blob | File | string,
+    type: string = "text/plain",
+  ): Promise<string> {
+    return lastValueFrom(this.readFile$(blob, type));
+  }
+
+  readFileAsDataURL$(
+    blob: Blob | File | string,
```
deltork, line 83: I like this!

packages/studio-web/src/app/file.service.spec.ts
```diff
@@ -25,4 +25,33 @@ describe("FileService", () => {
...
+  it("should create a data URL from a promise", async () => {
+    const got = await service.readFileAsDataURL("this is a test");
+    expect(got.startsWith("data:text/plain;base64")).toBeTrue();
+
+    // reverse the readFileAsDataURL operation.
+    const reversed = await fetch(got).then((resp) => resp.text());
+    expect(reversed).toBe("this is a test");
+  });
+
+  it("should turn utf8 to b64 and back", async () => {
```
deltork, line 43: Great!

packages/studio-web/src/app/shared/download/download.service.ts
```diff
@@ -137,13 +140,12 @@ Please host all assets on your server, include the font and package imports defi
         // Remove any images that are there from before
         currentPage.querySelectorAll("graphic").forEach((e) => e.remove());
         let graphic = doc.createElementNS(null, "graphic");
-        // @ts-ignore
-        let blob = await fetch(img).then((r) => r.blob());
-        blob = await compress(blob, 0.75);
+        let blob = await fetch(img)
```
deltork, line 143: This is a general question to discuss with the team, if we compress every time, won't image quality degradation be accelerated with each revision?

deltork (approved):
Looks good!

PR #452, opened 2025-08-06
refactor: replaced unsubscribe mechanism with DestroyRef

sergeleger:
### PR Goal? 

The `.pipe(takeUntil(this.unsubscribe$))` pattern requires boilerplate code to close all observables. 

Angular introduced `DestroyRef` (in v16) and `takeUntilDestroyed` (in v19):

```typescript
...
private destroyRef$ = inject(DestroyRef); 
...
interval(1000)
      .pipe(takeUntilDestroyed(this.destroyRef$))
      .subscribe((value) => {
        console.log(value);
      });
```

- `DestroyRef` [reference](https://angular.dev/api/core/DestroyRef)
- `takeUntilDestroyed` [reference](https://angular.dev/api/core/rxjs-interop/takeUntilDestroyed)

### Fixes? 

None, but two classes (demo and  editor components) did not properly implement `ngOnDestroy` boilerplate code.

### Feedback sought? 

Any comments or opinion on the proposed changed.

### Priority? 

Low.

### Tests added? 

None

### Confidence? 

High

### DIFF
--- a/packages/studio-web/src/app/demo/demo.component.ts
+++ b/packages/studio-web/src/app/demo/demo.component.ts
@@ -1,5 +1,3 @@
-import { Subject } from "rxjs";
-
 import { Component, OnDestroy, OnInit, ViewChild } from "@angular/core";
 import { Components } from "@readalongs/web-component/loader";
 
@@ -8,7 +6,7 @@ import { StudioService } from "../studio/studio.service";
 import { DownloadService } from "../shared/download/download.service";
 import { SupportedOutputs } from "../ras.service";
 import { ToastrService } from "ngx-toastr";
-import { WcStylingService } from "../shared/wc-styling/wc-styling.service";
+
 @Component({
   selector: "app-demo",
   templateUrl: "./demo.component.html",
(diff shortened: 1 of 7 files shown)


packages/studio-web/src/app/demo/demo.component.ts
```diff
@@ -18,7 +16,7 @@ import { WcStylingService } from "../shared/wc-styling/wc-styling.service";
 export class DemoComponent implements OnDestroy, OnInit {
   @ViewChild("readalong") readalong!: Components.ReadAlong;
   language: "eng" | "fra" | "spa" = "eng";
-  unsubscribe$ = new Subject<void>();
+
```
deltork, line 19: unsubscribe$ is unused so it makes sense to remove it

packages/studio-web/src/app/editor/editor.component.ts
```diff
@@ -56,7 +60,7 @@ export class EditorComponent implements OnDestroy, AfterViewInit {
   // a comma separated list of file extensions or mime types.
   htmlUploadAccepts = ".html";
 
-  unsubscribe$ = new Subject<void>();
+  private destroyRef$ = inject(DestroyRef);
```
deltork, line 63: I like this pattern it is more explicit

deltork (approved):
I prefer this pattern because it is more evident as to what is happening. Not sure if we have a way to stress test it.

PR #454, opened 2025-08-07
fix: reduced the sensitivity of view modification checks

### LINKED ISSUE #439: Studio-Web gets "dirty" too easily and warns about closing it unecessarily
If you click on the text box, or click on File in either the Text or the Audio section in step 1, the forms is already considered dirty and you get the warning that you might lose your data if you close or refresh the page. Same think if you just click on Take the tour without even going past the first tour step. Ditto opening the Editor without modifying anything in either the Editor or the Studio. Ditto if you open the Privacy dialog.

That warning is useful when there actually is data in the page, but it's annoying when you haven't actually put any data in, all you did was navigate trivially around it.

For the Studio side, we have tests that there is both text and audio data on Go to the next step, the same tests could be OR'd for considering the page dirty?

For the Editor side, I think you should have to have actually loaded a file before it's dirty.

And click on Privacy should never make the page dirty.

* I don't know if I'm using the word "dirty" correctly. In this Issue, I am using it to mean "causes the warning to appear on close or refresh".

sergeleger:
### PR Goal? 

More sensitive handling of the browser-generated confirmation dialog when navigating away or closing a Readalong instance.

For Studio: 

- fixes a `beforeunload` bug that always returned a isDirty=yes response. 
- added checks to verify if the language controls have been changed
- the event handler was not removed when the studio component was destroyed, this led to the Editor component inheriting the studio component isDirty logic.

For Editor:

- added new isDirty logic targeting the Editor component.

### Fixes? 

Closes #439 

### Feedback sought? 

Validation that the confirmation dialog only appears when it should (after meaningful user edits).

### Priority? 

Medium.

### Tests added? 

No.

### How to test? 

For Studio, any one of these should trigger the dialog box:

- typing text
- recording audio
- uploading a text or audio file
- changing the language from "Default" to "Specific"
- specifying a specific language

...and attempting to close or refresh the current tab.

For Editor:

- uploading a Offline HTML file.

...and attempting to close or refresh the current tab.

### Confidence? 

High

### DIFF
--- a/packages/studio-web/src/app/editor/editor.component.html
+++ b/packages/studio-web/src/app/editor/editor.component.html
@@ -42,7 +42,7 @@
           Offline-HTML file included in the Web Bundle download format.
         </p>
         <input
-          #textInputElement
+          #rasFileUpload
           (change)="onRasFileSelected($event)"
           class="form-control"
           name="text"
--- a/packages/studio-web/src/app/editor/editor.component.ts
+++ b/packages/studio-web/src/app/editor/editor.component.ts
@@ -39,12 +39,15 @@ import { WcStylingComponent } from "../shared/wc-styling/wc-styling.component";
   styleUrls: ["./editor.component.sass"],
   standalone: false,
 })
-export class EditorComponent implements OnDestroy, OnInit, AfterViewInit {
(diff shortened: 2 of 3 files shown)


deltork (approved):
Looks good!

PR #457, opened 2025-08-26
fix: text file delete button should not be visible when no file has been uploaded

### LINKED ISSUE #456: Writing text in the "Write" text box causes the Delete icon to appear under File
To reproduce:
 - Launch https://readalong-studio.mothertongues.org/dev/#/
 - Type some text under the Text box in Write mode
 - Click on File
 - see the "Delete" button: Image

Since no file was uploaded, the Delete button should not be there. Clicking it makes it disappear, but has no effect otherwise.

This behaviour does not happen in v1.6.0 (which is what is currently deployed in production), so it's due to a recent change.

sergeleger:
### PR Goal? 

Fixes bug introduced by #451, where the text file upload delete button was visible even if the user has not uploaded a text file.

### Fixes? 

Fixes #456 

### Feedback sought? 

That it now works correctly.

### Priority? 

Medium, bug fix and non-blocking.

### How to test? 

- Type some text under the Text box in Write mode
- Click on File
- The "Delete" button is not be visible.

### Confidence? 

High

### DIFF
--- a/packages/studio-web/src/app/upload/upload.component.html
+++ b/packages/studio-web/src/app/upload/upload.component.html
@@ -58,7 +58,7 @@
               <!-- </div> -->
             </div>
           </div>
-          @if (studioService.textControl$.value !== null) {
+          @if (textFileUpload.value) {
             <div class="row">
               <div class="col-12 offset-xl-8 col-xl-4">
                 <button


joanise (approved):
looks good, works correctly

PR #470, opened 2025-11-07
fix: read-along does not refresh on text changes

### LINKED ISSUE #469: Studio-Web doesn't update the text when you go back to step 1 and realign
When you go back to step 1 and realign after having edited the text, sometimes the text is still the old text.

One way reproduce:
 - load https://readalong-studio.mothertongues.org/#/ fresh
 - type `asdf` in the text, record a second of noise
 - select Upper Tanana as the language
 - go to next step
 - go back to step 1
 - change the text to `qwer`
 - go to next step
Result: a readalong is generated but it's still showing `asdf` as the text

Image

that should have shown `qwer`.

Actually, it looks like a simpler general recipe is:
 - fresh load
 - enter any text, audio
 - go to next step
 - back to step 1
 - change the text
 - go to next step
and the generated readalong has the original text, but if you continue looping, going back between step 1 and next step changed text is picked up, so it seems like it's only the first time you go back that something wonky happens.

sergeleger:
### PR Goal? 

Fixes bug where the `<read-along />` component in Step 2, displays the incorrect version of the text after a user edits the text.

### Fixes? 

Fixes #469 

### Feedback sought? 

That the bug has been fixed.

### Priority? 

Medium, non-blocker.

### Tests added? 
No.

### How to test? 

- fresh load
- enter any text, audio
- go to next step
- back to step 1
- change the text
- go to next step
- The text in the Read-Along should have been updated also.

### Confidence? 

High, but I'm not convinced this is the best approach.

### DIFF
--- a/packages/studio-web/src/app/demo/demo.component.ts
+++ b/packages/studio-web/src/app/demo/demo.component.ts
@@ -38,6 +38,9 @@ export class DemoComponent implements OnDestroy {
 
     this.studioService.b64Inputs$.subscribe(async (b64Input) => {
       if (b64Input[1]) {
+        // disable the read-along component until the
+        // `await` below is done and updates the dataUrl.
+        this.rasAsDataURL.set("");
         this.rasAsDataURL.set(await this.b64Service.rasToDataURL(b64Input[1]));
       }
     });


joanise:
Thanks for the quick fix! Yes, it fixes the bug. Did I mention I hate JS and rx and all that stuff? It all seems so fragile!

joanise (approved):
Looks so trivial, yet can be tricky to figure out. Works, fixes the bug.

Issue #462, opened 2025-09-16
If you save web-component preferences, you cannot undo it, and it affects the Studio-Web


joanise:
If you change the web-component settings using the gear menu, e.g., the display language, you cannot unsave them, and they affect Studio-Web in all languages.

Jarring workflow example:
 - Go to https://readalong-studio.mothertongues.org/
 - create a readalong
 - click on the gear and ask for Spanish
 - click on "Remember my preferences for this site"
 - Go to https://readalong-studio.mothertongues.org/fr
 - create a readalong
 - ❗   the web component interface is in Spanish 

The only way to get back to default behaviour (where the Studio-Web displays the web-c in the studio language) is to clear browsing data for the site, in which case you also lose your privacy opt-out or opt-in setting for analytics.

Possibilities:
 - Add a button like "Clear saved preferences" that is active if you have saved preferences? This button would remove those saved preferences that were saved via the gear menu.
 - have the Studio-Web language override the web-component display language? not sure I like this, but it would make the Studio-Web behave more coherently. This might mean the saved preferences would apply only when a web component is loaded outside the studio, which might be confusing too, not sure.

My preference is to have a "Clear saved preferences" button.

sergeleger:
Should clearing preferences do both: (a) remove the settings from local storage and (b) reset the preferences to the default values?

In some cases the "defaults" will no longer available: user preferences are available as component properties and those should be the default: `<read-along language="fra" />`. Language defaults to "eng" if the property is not provided, the user's preference is "fra", the setting's menu overwrites this `Prop` value and there's no way to recover the original value.

To properly implement this functionality, we will need to  refactor each read-write `Prop` to a read-only `Prop` and introduce a second read-write `@State` variable.

joanise:
I'm not sure I understand the subtleties of your question, but certainly yes to a), the key is to remove the settings from local storage. I'm not sure what b means, but I think we want to not have preferences, and therefore let default values apply. Am I just saying the same thing in different words?

I don't know if refactoring is needed: if you never saved any preferences, the default or the HTML attribute are used when you load a readalong. Would just clearing the local storage not restore that behaviour? But if you think refactoring is the way to go, then go for it.

In any case, as long is the user is able to get back to the state you're in before you're ever touched settings, this issue will be solved.

sergeleger:
There are three sources for determining the behaviour of the component:

1. the configuration done via the settings window, and tentatively stored in local storage: `language=fra`
[...]

PR #472, opened 2025-11-14
feat: added functionality to remove saved user preferences

sergeleger:
### PR Goal? 

Removing stored user preferences required:

> to clear browsing data for the site, in which case you also lose your privacy opt-out or opt-in setting for analytics.

this PR adds a "Reset" button to the settings panel that will reset all configurations (including deleting the stored settings).

### Fixes? 

#462 

### Feedback sought? 

- implementation works as intended.
- feedback on UI elements in the settings page (new Reset button, and new location of "save preferences" button)
- feedback on new En / Fr / Sp messages in `i18n/` folder.

Also, additional minor improvements to the settings modal window:

- version paragraph had a cursor pointer (giving the impression that it could be clicked) - removed
- added dark mode CSS for the language select box, the previous implementation had very low contrast (black text on blue background)
- the escape key did not always close the settings window.
   - a quick test: click on an empty section of the settings window: escape would not work 
- language names in select box are now displayed in their native language (English, Francais, Espanol) regardless of the user's current language choice.

### Priority? 

Medium, non blocker.

### Tests added? 

Yes

### How to test? 

Documented steps from the issue #472:

- Go to https://readalongs.github.io/Studio-Web/pr-preview/pr-472/
- create a readalong
- click on the gear and ask for Spanish
- click on "Remember my preferences for this site"
- Go to https://readalong-studio.mothertongues.org/pr-preview/pr-472/fr/#/
- create a readalong
-❗ the web component interface is in Spanish
- click on the gear and click on the reset button
- **the web component should be in French.**

Other:

- Reset can be used even if no settings have been saved, an easy way to set the defaults.

### Confidence? 

High.


packages/web-component/src/i18n/messages.spa.json
```diff
@@ -23,7 +23,9 @@
     "page-animation": "Transiciónes de página animadas",
     "auto-pause": "Pausa al final de cada página",
     "language": "Idioma de la interfaz",
-    "save-settings": "Recordar mis preferencias para este sitio",
+    "save-settings": "Recordar mis preferencias",
+    "reset-title": "Restablecer todas las preferencias",
+    "reset": "Reiniciar",
```
joanise: I think "Reinicializar" would be better.

joanise: other strings look good to me

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -99,15 +115,19 @@ export class ReadAlongComponent {
...
   /**
-   * Theme to use: ['light', 'dark'] defaults to 'dark'
+   * Theme to use: ['light', 'dark'] defaults to the user's configured theme.
    */
-  @Prop({ mutable: true, reflect: true }) theme: string = "light";
+  @Prop() theme: string = navigatorTheme();
+
+  @State() private userPreferences: UserPreferences;
+  @State() private userPreferencesDirty: boolean = false;
+  @State() private hasUserPreferences: boolean = hasUserPreferences();
```
joanise: Is there any reason not to place these three states with the other ones in the States section below?

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -154,8 +174,7 @@ export class ReadAlongComponent {
    * animated, good for fast computers) or "auto" (choppy but much less compute
    * intensive)
    */
-  @Prop({ mutable: true, reflect: true }) scrollBehaviour: ScrollBehaviour =
-    "smooth";
+  @Prop() scrollBehaviour: ScrollBehaviour = "smooth";
```
joanise, line 173: Took me a while to figure out why this was still used -- it appears to just be the thing that sets the default. Is this also what gets modified if you set it in the HTML?

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1032,21 +1051,28 @@ export class ReadAlongComponent {
...
     //this.cssUrl = this.urlTransform(this.cssUrl)
+
     // TO maintain backwards compatibility language code
-    if (this.language.length < 3) {
-      if (this.language.match("fr") != null) {
-        this.language = "fra";
-      } else if (this.language.match("es") !== null) {
-        this.language = "spa";
+    if (prefs.language.length < 3) {
+      if (prefs.language.match("fr") != null) {
```
joanise: This is weird code. Why not `prefs.language !== "fr"`? @deltork ? Is this case insensitive?

joanise (commented):
Works well when I test it.
I have some questions, see below.

I didn't dig enough to understand why most instances of `this.X` had to be replaced by `this.userPreferences.X` but it works.

I didn't test if we can still change settings in the html as attributes to the `read-along` element and have them reflected when you load a readalong. Did you exercise that?

Finally, I'm leaving comments but I'm not confident enough to really review, @deltork I'm hoping you'll do a code review too.

packages/web-component/src/components/read-along-component/read-along.tsx
sergeleger: No reason, moved them here is early implementation work. Returned to the States section below.

packages/web-component/src/components/read-along-component/read-along.tsx
sergeleger: Yes, it contains the value passed in the HTML `<read-along scroll-behaviour="..." />`, if this attribute (or property) is not used, it defaults to "smooth".

Answering the question about `this.X` vs `this.userPreferences.X` here also: `this.X` contains the initial state of the component, either the hard-coded value or the attribute value passed in `<read-along />`. The previous implementation directly modified `this.X` overwriting the component's initial state. This refactor makes `this.X` read-only and it is used only for setting the component's initial state and clearing the users's preferences (see: `defaultUserPreferences()` where `this.X` is used) .

sergeleger:
> I didn't test if we can still change settings in the html as attributes to the `read-along` element and have them reflected when you load a readalong. Did you exercise that?

If I'm understanding the use-case correctly something along the lines of?

```javascript
const el = document.getElementById("readalongInstance");
el.scrollBehaviour = "auto";
```
Currently would not work. Neither would this:

```javascript
const el = document.getElementById("readalongInstance");
if(el.scrollBehaviour == "auto") {
...
}
```

In the first case it would take effect only when/if the user performs a "Reset". In the second, the value returned would always reflect the component's initial state and not the current user defined preference. Both can be addressed. 

Is there an example of this or a test?

joanise:
What I meant by "setting the attributes" is right in the HTML itself, someone publishing a readalongs could set `language="fra"` or the scrolling parameter etc. I think all the use cases are covered by examples in `packages/web-component/test-data/ej-fra/*.html` so I suppose if CI exercises those, then they're already tested.

joanise:
Thanks for the changes. Looks good to me, but I'll let Del approve once he's done his review.

packages/web-component/src/components/read-along-component/read-along.tsx
deltork: This allows the app to support both ISO **639-3** and ISO **639-1** (`"fra"` and `"fr"`) for maximum compatibility.

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -99,15 +115,15 @@ export class ReadAlongComponent {
   @Prop() svgOverlay: string;
 
   /**
-   * Theme to use: ['light', 'dark'] defaults to 'dark'
+   * Theme to use: ['light', 'dark'] defaults to the user's configured theme.
    */
-  @Prop({ mutable: true, reflect: true }) theme: string = "light";
+  @Prop() theme: string = navigatorTheme();
```
deltork: I like this, using the device theme as default!

sergeleger:
Yes, that has been implemented and manually tested. Values passed in HTML become the component's default values when "Reset" is used.

packages/web-component/src/components/read-along-component/read-along.tsx
deltork: This makes it case insensitive
```suggestion
      if (prefs.language.match(/fr/i) != null) {
```

packages/web-component/src/components/read-along-component/read-along.tsx
```diff
@@ -1032,21 +1051,28 @@ export class ReadAlongComponent {
...
     // TO maintain backwards compatibility language code
-    if (this.language.length < 3) {
-      if (this.language.match("fr") != null) {
-        this.language = "fra";
-      } else if (this.language.match("es") !== null) {
-        this.language = "spa";
+    if (prefs.language.length < 3) {
+      if (prefs.language.match("fr") != null) {
+        prefs.language = "fra";
+      } else if (prefs.language.match("es") !== null) {
```
deltork:
```suggestion
      } else if (prefs.language.match(/es/i) !== null) {
```

deltork:
I have a minor tweak for the settings UI. see #475.
Otherwise everything looks good to me.

deltork (approved):
Please review and merge #475 into this

packages/web-component/src/components/read-along-component/read-along.tsx
joanise: I see some `!=` with just one equal in there, should they not all be `!==`?

sergeleger: Could we convert this to a data structure? Regex feels like an over kill and in this case not specific enough (`abc_fr_abc` would be accepted for French):

```
const acceptedLanguages = {
"en": "eng",
"eng": "eng",
"fr": "fra",
"fra": "fra",
"es": "spa",
"spa": "spa",
}
prefs.language = acceptedLanguages[prefs.language.toLowerCase()]||"eng";
```

joanise: That's not quite accurate since the code is in a if block where len<3... which also points to overkill. :) 

I like your map, though, seems like a good idea. In Python, [] with an unfound value would raise an exception, here would that return a falsy result instead?

deltork: > I see some `!=` with just one equal in there, should they not all be `!==`?

yeh they should be `!==`

sergeleger: > That's not quite accurate since the code is in a if block where len<3... which also points to overkill. :)
 
Oops, I assumed there was an ` else ` clause hiding in all the short diff blocks. Since there are none, there's no 3-character language validation either. But `getI18nString` handles the bad language input.

> 
> I like your map, though, seems like a good idea. In Python, [] with an unfound value would raise an exception, here would that return a falsy result instead?

That's correct, and the ` || "eng"`  returns English in those cases (it does not throw an exception).

sergeleger: I'll update the language validation code after Del merges PR #475  (which I've just accepted).

deltork (approved):
Looks good to me

PR #483, opened 2025-12-10
Dev.ej/better g2p toast

joanise:
### PR Goal? 

 - make the g2p error toast stay longer
 - enable copy-pasting from the g2p error toast by neither dismissing it on click nor on hover
 - clear all outstanding toasts on successful alignment, they're no longer relevant once we get to step 2
 - disable spell checking on the text area

### Feedback sought? 

Whether the updated app feels better

### Priority? 

medium

### Tests added? 

no

### How to test? 

enter `asdf 123 asdf : 1234` in the text box, and see:
 - `asdf` is no longer underlined as a spelling error
 - when you click on go to next step, you're able to interact with the g2p error toast without making go away
 - when you fix your text and align again, all the no-longer relevant toasts are gone

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/packages/studio-web/project.json
+++ b/packages/studio-web/project.json
@@ -82,7 +82,7 @@
       "defaultConfiguration": "production"
     },
     "update-browserslist": {
-      "command": "npx update-browserslist-db@latest"
+      "command": "npx -y update-browserslist-db@latest"
     },
     "serve": {
       "executor": "@angular-builders/custom-webpack:dev-server",
--- a/packages/studio-web/src/app/upload/upload.component.ts
+++ b/packages/studio-web/src/app/upload/upload.component.ts
@@ -584,6 +584,7 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
         next: (progress) => {
           if (progress.hypseg !== undefined) {
             this.loading = false;
+            this.toastr.clear(); // clean all outstanding toasts on alignment success
             this.stepChange.emit([
               "aligned",
               this.studioService.audioControl$.value,
--- a/packages/studio-web/src/app/upload/upload.component.ts
+++ b/packages/studio-web/src/app/upload/upload.component.ts
@@ -176,7 +176,10 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
         );
       }
       this.toastr.error(err.error.detail, $localize`Text processing failed.`, {
-        timeOut: 30000,
+        timeOut: 180000,
+        closeButton: true,
+        tapToDismiss: false,
+        disableTimeOut: "extendedTimeOut",
       });
     } else {
       this.toastr.error(
--- a/packages/studio-web/src/app/app.module.ts
+++ b/packages/studio-web/src/app/app.module.ts
@@ -40,7 +40,11 @@ defineCustomElements();
     BrowserModule,
     AppRoutingModule,
     BrowserAnimationsModule,
-    ToastrModule.forRoot(),
+    ToastrModule.forRoot({
+      preventDuplicates: true,
+      resetTimeoutOnDuplicate: true,
+      includeTitleDuplicates: true,
+    }),
     ReactiveFormsModule,
     MaterialModule,
     MatToolbarModule,
--- a/packages/studio-web/src/app/upload/upload.component.html
+++ b/packages/studio-web/src/app/upload/upload.component.html
@@ -116,6 +116,7 @@
                   class="border rounded b-0 p-0 bg-light"
                   rows="8"
                   matInput
+                  spellcheck="false"
                   i18n-placeholder="Example input text"
                   placeholder="Ex. Hello my name is...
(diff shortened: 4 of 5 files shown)


marctessier (approved):
I got this toast at first below and I had the time  to copy it into this message :-)

```
Text processing failed.
g2p could not be performed, please check your text or your language code. These words could not be converted from text to phonemes by g2p: ' 123 ', ' : ', ' 1234 '.
```

`ASDF` did not have a red squiggly red line under it in the text area  :-) 

After converting  the numbers to words ex: 1 ==> one

It was still complaining about the ":"  

```
Text processing failed.
g2p could not be performed, please check your text or your language code. These words could not be converted from text to phonemes by g2p: ' : '.
```
 
So I changed the language to  "English" and it processed to  the next step. 

NOTE: =>  I did see a red and yellow/orange toast fly by the screen  on the first try "very quickly" before going to step 2.. 

I clicked on step one to try that again... I re-click on next step and this time only the yellow/orange toast  "`with some content`"   flash on the screen... ( Unable to read what it's about / I tried a few times ;-) but it's a warning and we are probably better of going to the next step since it was a "Congratulations! Here's your ReadAlong!"

joanise:
> NOTE: => I did see a red and yellow/orange toast fly by the screen on the first try "very quickly" before going to step 2..

Oh, hum, that's a poor side effect of my change to dismiss toasts once you're on step 2. It's probably the alignment failure toasts, telling you it has to try again with relaxed parameters. I wasn't thinking about them flashing like that, but what you describe is not good. I'll have to think about a better solution than that... Thanks for flagging!

packages/studio-web/project.json
```diff
@@ -82,7 +82,7 @@
       "defaultConfiguration": "production"
     },
     "update-browserslist": {
-      "command": "npx update-browserslist-db@latest"
+      "command": "npx -y update-browserslist-db@latest"
```
sergeleger, line 85: :100: 

I actually took note of this issue today!  Glad it's fixed.

packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -176,7 +176,10 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
         );
       }
       this.toastr.error(err.error.detail, $localize`Text processing failed.`, {
```
sergeleger, line 178: Something like this might help persist the alignment warnings:

```javascript
this.textProcessingErrToast = this.toastr.error( ... ).toastId;
```

packages/studio-web/src/app/upload/upload.component.ts
```diff
@@ -581,6 +584,7 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
         next: (progress) => {
           if (progress.hypseg !== undefined) {
             this.loading = false;
+            this.toastr.clear(); // clean all outstanding toasts on alignment success
```
sergeleger, line 587: and clear only the long-lived text processing error message.

```javascript
            if (this.textProcessingErrToast) {
              this.toastr.clear(this.textProcessingErrToast); // clean all outstanding toasts on alignment success
              this.textProcessingErrToast = 0;
            }
```

sergeleger:
> > NOTE: => I did see a red and yellow/orange toast fly by the screen on the first try "very quickly" before going to step 2..
> 
> Oh, hum, that's a poor side effect of my change to dismiss toasts once you're on step 2. It's probably the alignment failure toasts, telling you it has to try again with relaxed parameters. I wasn't thinking about them flashing like that, but what you describe is not good. I'll have to think about a better solution than that... Thanks for flagging!

Here's the warning message I'm getting:
image

packages/studio-web/src/app/upload/upload.component.ts
joanise: Nice, thanks for the suggestion on how to do this.

joanise:
> > > NOTE: => I did see a red and yellow/orange toast fly by the screen on the first try "very quickly" before going to step 2..
> > 
> > 
> > Oh, hum, that's a poor side effect of my change to dismiss toasts once you're on step 2. It's probably the alignment failure toasts, telling you it has to try again with relaxed parameters. I wasn't thinking about them flashing like that, but what you describe is not good. I'll have to think about a better solution than that... Thanks for flagging!
> 
> Here's the warning message I'm getting: image

Oh, right, just processed this carefully: the problem is with English g2p: OOVs get mapped to nothing. So "This asdf test" gets g2p'd to the same thing as "This test", and we end up losing a word in the process. Instead of dropping the word, we map it using the "und" universal g2p mapping. This is expected given your input, and not a bug. And I guess it would be important not to hide it, the user might want to be able to read this.

joanise:
Closing this without merging: I'm going to turn this into three separate PRs to merge independently, since one of the changes combined here has problems and isn't ready to go.

PR #484, opened 2025-12-10
fix: better messages when alignment fails

joanise:
### PR Goal? 

When we get to the third alignment failure toast, the message was confusing and not helpful, so improve it.
When we get to the second alignment failure toast, it should be a warning only since we automatically try again.

### Fixes? 

The note quite intuitive messages we get on alignment failure.

Before | After
--- | ---
image | image

### Feedback sought? 

Do you agree this is more intuitive?

### Priority? 

low

### Tests added? 

no

### How to test? 

enter this as text:
```
This is a test, a very long test, and it cannot possibly be aligned to a second or two of noise, so alignment will eventually fail, with three distinct toasts popping up along the way.
```
record a second or two of noise

then click on go to the next step

### Confidence? 

medium high

### Version change? 

no

### DIFF
--- a/packages/studio-web/src/app/upload/upload.component.ts
+++ b/packages/studio-web/src/app/upload/upload.component.ts
@@ -189,8 +189,7 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
 
   reportUnpronounceableError(err: Error) {
     this.toastr.error(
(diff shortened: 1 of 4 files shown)


sergeleger (approved):
This works for me and the warning toasts for first and second failures are logical since the audio processing has not yet fully completed.

PR #489, opened 2025-12-15
Dismiss most toasts on successful alignment

joanise:
### PR Goal? 

This is a redo on my first naive idea if dismissing all toasts on successful alignment. Marc and Serge noticed toasts would pop and go without being able to read them.

With this PR:
 - I am dismissing remaining success toasts, because those would have been from recording audio or the like
 - I am dismissing remaining error toasts, because if alignment succeeds to have to have handled those problems
 - I am keeping warning toasts because you probably still want to read them, especially if they were issues by the alignment process
 - In case there's an alignment relaxing warning, I'm adding a success toast with recommendations to review the results
 - And I'm disabling dismiss toasts on hover globally, because that behaviour just really irks me -- if I hover on a toast, it's usually because I'm paying attention to it, not because I want it gone!

### Fixes? 

The motivation here is that when you get to step 2, many of the toasts are really not relevant anyway, but they're annoyingly hiding part of the UI, so I'm trying to fix that.

### Feedback sought? 

Review the new messages and, Serge or Marc, their French translation.

### Priority? 

pretty high because I want to release a new patch for S-Web to roll in a number of bug fixes we've made, and if we're happy with this PR, I'd like to include it. Mind you, if you see problems with this PR I will just release without it, so it's only high priority to tell me "yes" or "not sure", elaborating on "not sure" would be less urgent.

### Tests added? 

no, but I adjusted existing tests to work with these changes.

### How to test? 

1.
 - Enter mismatched audio and text, e.g., `asdf asdf asdf asdf asdf` against a second or two of silence/noise
 - go to next step
 - See one or two warning toasts, then when you get to step 2 the "Audio recorded" toast is gone, but you still see the warning(s) and a success with caveat toast

2.
 - Enter `test asdf test` and chose English, and whatever audio
 - go to next step
 - The warning that `asdf` needed to fall back to und is still visible, and possibly toast about alignment

3.
 - Enter `asdf 42 asdf : asdf` and whatever audio
 - go to next step
 - See the long-last copy-pasteable g2p error toast
 - Fix the text, e.g., to `asdf asdf asdf`
 - go to next step
 - The long-lasting g2p error toast is gone and you only see alignment-related toasts, if any toasts at all

### Confidence? 

medium-high

### Version change? 

not for this, but yes, I want to release a patch bump


packages/studio-web/tests/test-commands.ts
```diff
@@ -107,10 +107,17 @@ export const testMakeAReadAlong = async (page: Page) => {
...
+      .locator("div.toast-message")
+      .count()
+      .then(async (count) => {
+        if (count > 0) {
+          await page.locator("div.toast-message").last().click({ force: true });
+          await expect(async () => {
+            await expect(page.locator("div.toast-message")).not.toBeVisible();
+          }).toPass();
+        }
+      });
```
deltork:
```suggestion
    for(const toast of await page.locator("div.toast-message").all()){
        await toast.click({ force: true });
    }
    await expect(async () => {
      await expect(page.locator("div.toast-message")).not.toBeVisible();
    }).toPass();

joanise:
MT feedback:
 - README.md needs work, the instructions don't quite work as is -- EJ: will fix in a separate PR
 - The g2p warning toast where it's about fallback to und is not helpful to 99% of users. Only the super user might want to see it. EJ idea -- let's hide it by `console.log()`ging it rather than issueing a toast with it.
 - it would be nice to hyperlink the app version tag at the bottom to our GitHub page

deltork (changes requested):

PR #494, opened 2025-12-19
perf: disable shards processing in CI

joanise:
There is too much overhead from npm install, it cancels all benefits from sharding

### PR Goal? 

We had end-to-end tests parallelized in three jobs, and merged after, but it turns out the overhead of the npm install required in each subjob makes the whole thing slower parallelized than in a single job.

Split example: 
https://github.com/ReadAlongs/Studio-Web/actions/runs/20348658434
shards took 3:49, 3:56, 3:33, merge took 1:11
time elapsed = 3:45 + 1:11 = 4:56
CI minutes used = sum = 12:29

Single job example:
https://github.com/ReadAlongs/Studio-Web/actions/runs/20374155875
non-parallel took 4:13 = time elapsed = CI minutes used

With my apologies for undoing this nice parallelization.

### Feedback sought? 

sanity check

### Priority? 

normal

### Confidence? 

high

### DIFF
--- a/.github/workflows/end-to-end-tests.yml
+++ b/.github/workflows/end-to-end-tests.yml
@@ -74,76 +74,46 @@ jobs:
     name: Studio Web test-suites
     timeout-minutes: 60
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        shardIndex: [1, 2, 3]
-        shardTotal: [3]
     steps:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
+
       - uses: actions/setup-node@v6
         with:
           node-version: lts/*
-      - name: Install and run the back-end API, needed for end-to-end testing
+
+      - name: Install and run the back-end API
         run: |
           git clone https://github.com/ReadAlongs/Studio
           cd Studio
           pip install -e .[api]
           ./run-web-api.sh &
-          # wait for the API to be up
-          curl --retry 20 --retry-delay 1 --retry-all-errors http://localhost:8000/api/v1/langs
+
       - name: Install everything
         run: npm ci
+
       - name: Run studio-web in the background
         run: |
           npx nx build web-component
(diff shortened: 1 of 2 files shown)


.github/workflows/end-to-end-tests.yml
```diff
@@ -74,76 +74,46 @@ jobs:
...
-      - name: Install and run the back-end API, needed for end-to-end testing
+
+      - name: Install and run the back-end API
         run: |
           git clone https://github.com/ReadAlongs/Studio
           cd Studio
           pip install -e .[api]
           ./run-web-api.sh &
-          # wait for the API to be up
-          curl --retry 20 --retry-delay 1 --retry-all-errors http://localhost:8000/api/v1/langs
```
deltork, line 96: This removal I assume is necessary? (sharding was issue?)

.github/workflows/end-to-end-tests.yml
```diff
@@ -74,76 +74,46 @@ jobs:
...
           npx nx run-many --targets=serve,serve-fr,serve-es --projects=web-component,studio-web --parallel 6 &
 
+      - name: Wait for the servers to be up
+        run: |
           # wait for the studio web to be up
           sleep 50
           curl --retry 20 --retry-delay 10 --retry-all-errors http://localhost:4200 > /dev/null
+          # Make sure the back-end API is up too
+          curl --retry 20 --retry-delay 1 --retry-all-errors http://localhost:8000/api/v1/langs
+
```
deltork, line 108: Ah! good place to put it! Ignore the previous comment

deltork (approved):
Looks good

.github/workflows/end-to-end-tests.yml
joanise: exactly, I thought it better to defer the check since it's silly to wait for it before starting the npm ci

PR #496, opened 2025-12-19
fix: hide g2p warnings from the GUI, show them in console

joanise:
### PR Goal? 

Marc noticed that sometimes our toast about g2p warnings can get really long, especially when aligning a long text, but it's not actually informative. 99% of users won't care that this or that word had to fall back to `und` or even understand what that means.

In brainstorming, Marc and I figured it would make more sense to `console.log()` the RAS web api log for the advanced user who wants to see them, but keep them out of the GUI because they're just visual noise to most users.

So that's what this PR does

### Feedback sought? 

I guess vetting that this idea makes sense.

### Priority? 

normal

### Tests added? 

nope

### How to test? 

 - Enter `test asdf test` in the text
 - any audio
 - select English as the language
 - and click "go to the next step"
 - See that the toast about g2p failing with `eng` for `asdf` is not diplayed, but if you open the console you'll see the full assemble log including the g2p warnings with header `RAS Web API assemble/ log:`.

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/packages/studio-web/src/app/upload/upload.component.ts
+++ b/packages/studio-web/src/app/upload/upload.component.ts
@@ -533,16 +533,8 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
     })
       .pipe(
         switchMap(({ audio, ras }: { audio: IAudioBuffer; ras: ReadAlong }) => {
-          if (ras.log !== null) {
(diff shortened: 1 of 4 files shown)


marctessier (approved):
Works great!  gp2  Warning were in console logs.

PR #499, opened 2025-12-22
fix: make npx nx bundle web component automatically build first

### LINKED ISSUE #445: Make `npx nx bundle web-component` automatically build first
Bit gotcha at the moment: if you bundle without having first done build, you just re-bundle what was already in `dist/`, which might be stale.

For safer procedure, make `npx nx bundle web-component` depend on build so it's just automatically done.

sergeleger:
### PR Goal? 

Forces a web component build when the bundle script is called, this avoids bundling a stale version of the readalong web component.

Also, since `npx nx bundle web-component` now performs a build operation, simplified the CI scripts to avoid unnecessary builds. 

### Fixes? 

Closes #445 

### Priority? 

Low

### Tests added? 

No

### How to test? 

Not easy to test: one approach would be to 

* delete the `dist/packages/web-component/` directory. 
* `npx nx bundle web-component` succeeds (while on `main` it would not)

### Confidence? 

High

### Version change? 

No

### DIFF
--- a/.github/workflows/bundle.yml
+++ b/.github/workflows/bundle.yml
@@ -18,7 +18,6 @@ jobs:
           package-manager-cache: false
       - run: npm install
       - run: npx update-browserslist-db@latest
-      - run: npx nx build web-component
       - run: npx nx bundle web-component
       - name: Create Pull Request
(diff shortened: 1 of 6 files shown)


joanise (approved):
That's a great idea, thanks! We should have done this a long time ago.

PR #501, opened 2025-12-23
fix: tall image clipping on read alongs

### LINKED ISSUE #468: Title page image style needs to be adjusted
I have reports of the images on the title page getting cropped. Possible CSS issue.

Image
the full image
Image

it looks ok outside of the title page

sergeleger:
### PR Goal? 

This PR fixes the clipping of tall images, which sometimes occurs when text column of the read along is narrow.

The applied changes are to properly implement HTML nesting rules and CSS Grid rules which lets the browser  infer width and height information on the parent container of the image.

* HTML nesting rules states that inline elements like `<span />` cannot contain block elements like `<div />`. The `` tag had that hierarchy (`html > body > ... > div > span > div > img`)

* CSS Grid rules requires that grid cells must be defined directly underneath the grid parent container, here's an incorrect implementation:

```html
<div style="display: grid">
   <div> <!-- introduces a layer between grid parent and child -->
         <div style="grid-area: image"> ... </div>
   </div>
</div>
```

Correcting both issues seems to have corrected the tall image clipping problem.

### Fixes? 

Closes #468

### Feedback sought? 

Validate my assumptions. I'm also not familiar with other ReadAlongs UI design decisions that this solution breaks.

### Priority? 

Medium, but it is a bug reported by users.

### Tests added? 

No, but tests still passes after restructuring the HTML.

### How to test? 

To recreate the broken behavior (in production for example):

* create a new read along with a single short sentence (e.g.: this is a test)
* upload the image shared by Del (in issue #428)
* notice the bottom of the image is truncated
* add a new translation for "this is a test" and type in a lot of translation text forcing the text grid to get wider and the image will eventually resize so that it is no longer clipped.

With this fix:

* create a new read along with a single short sentence (e.g.: this is a test)
* upload the image shared by Del (in issue #428)
* **Image is no longer clipped**

### Confidence? 

Medium. As mentioned above, this seems to work but have I introduced breaking UI look and feel.


deltork (approved):
Code looks good to me!
You can move the icon to the top left by setting the margin of `#img-remover` to `0`

PR #503, opened 2025-12-23
revert: reimplement read alongs's getImages() API

### LINKED ISSUE #502: A removed image from a read along is still included in Offline HTML exports
Removing an image from a read along and creating an Offline HTML will _sometimes_ include the removed image.

These steps normally reproduces the problem:

* create a read along
* add an image
* **create HTML export**
* remove the image
* create HTML export and the image is still included.

I'm not certain the bolded step is required.

sergeleger:
### PR Goal? 

Reverts an inadvertent API  change to the web component's `getImages()` functionality. The correct API should return an object with `null` values for images that had been deleted from the read along:

```javascript
console.log(readalong.getImages()); 
// {"1": "blob://...", "2": null}
```
Studio Web uses this `null` information as an indicator to remove a previously attached image from the download package.

### Fixes? 

Closes #502 

### Feedback sought? 

Still works as expected.

### Priority? 

Low.

### Tests added? 

No.

### How to test? 

* create a read along
* add an image
* create HTML export
* remove the image
* re-create HTML export 
* open the latest exported read along and image is  no longer included.

### Confidence? 

High.

### Version change? 

No.

### DIFF
--- a/packages/web-component/src/components/read-along-component/read-along.tsx
+++ b/packages/web-component/src/components/read-along-component/read-along.tsx
@@ -1472,8 +1472,10 @@ export class ReadAlongComponent {
       URL.revokeObjectURL(this.images[pageIndex]);
     }
 
-    delete this.images[pageIndex];
-    this.images = { ...this.images }; // Using spread operator as advised https://stenciljs.com/docs/reactive-data#updating-an-object
+    const newImage = {};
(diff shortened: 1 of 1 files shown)


joanise (approved):
Sorry, I missed this one. All good, thanks for fixing this!

PR #508, opened 2026-01-12
build(deps): bump angular/core and /compiler to 20.3.16

joanise:
bump angular/core and /compiler to 20.3.16 and other peers that have to come along, plus rebuild a fresh package-lock.json because otherwise the npm install just fails.

Replaces #505 and #506

### PR Goal? 

Handle CVEs in @angular/core and @angular/compiler <= 20.3.15

### Fixes? 

Those CVEs

### Feedback sought? 

rubber stamping, as CI's output is what I mostly want

### Priority? 

High according to the dependabot alerts...

### Tests added? 

no

### How to test? 

I'm assuming if the PR preview runs, that's good enough.

### Confidence? 

medium-high

### Version change? 

no


joanise:
Meh, this installed on my machine using node 22, but it's not installing here in CI with node 24. I guess I'll have to test node 24 locally.

sergeleger (approved):
Looks good to me.

PR #526, opened 2026-03-02
fix(deps): address several vulnerabilities via npm audit fix

joanise:
### PR Goal? 

auto fix vulnerabilities that can be auto fixed


joanise:
This PR breaks testing. We can't bump all these packages until we replace karma by something else.

PR #532, opened 2026-03-06
fix: improvement the privacy message in all languages

joanise:
### PR Goal? 
 
- the fact that we have to pay for Playsible is not relevant, that's our problem
- "which devices" feels creepy to me, like we're tracking indivual devices. I already worded it as types of devices in French and Spanish when I translated, and here I am backporting that fix to English too.

### Feedback sought? 

Do you like the new text?

To see the differences more clearly, use the SemanticDiff link, there are really just small changes which the github diff fails to highlight well.

### Priority? 

low

### How to test? 

look at the PR preview

### Confidence? 

high

### Version change? 

no


joanise:
Oh, how disappointing! SemantifDiff did not pick up on the fact that the massive line in French and Spanish has just a few small changes.

dlothian:
In the translation you gave me for the WW blurb you included "(en anglais)" but not in this one. Should it be here too?

joanise:
> In the translation you gave me for the WW blurb you included "(en anglais)" but not in this one. Should it be here too?

You know, it's not really necessary, and it takes more space for little value.

PR #548, opened 2026-05-08
fix: provide a specific error message when recording permission is denied

### LINKED ISSUE #149: Need to check microphone permissions on page load
If ever you deny microphone permissions to a page, the browser (any sensibly secure one) will never ask you again, which means that recording will always fail with permission denied.

We need to detect this situation and warn the user right away, that they will need to grant permissions and reload the page.

We also need to delay the "recording" indicator until the microphone actually record when the browser asks for permission.

joanise:
### PR Goal? 

This is work I did a long time ago, but I forgot to PR it...

If microphone permission is denied, this PR shows:
image

Before this PR:
image

### Fixes? 

Fixes #149 
Well, really, it finishes fixing that issue by adding a friendlier user message when microphone permission is denied; the other half of the issue was done in #154, and all this is work that builds on top of #127.

### How to test? 

Open the PR preview and the prod version in an incognito window, and refuse microphone access when prompted.

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/packages/studio-web/src/app/upload/upload.component.ts
+++ b/packages/studio-web/src/app/upload/upload.component.ts
@@ -281,7 +281,15 @@ Please check it to make sure all words are spelled out completely, e.g. write "4
       await this.microphoneService.startRecording();
       this.recording = true;
     } catch (err: any) {
-      this.toastr.error(err.toString(), $localize`Could not start recording!`);
+      let message;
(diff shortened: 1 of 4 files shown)


roedoejet (approved):
yep - this is a good change, thanks @joanise

PR #551, opened 2026-05-11
fix: allow 400kb .readalong XML file, for a particular project

joanise:


roedoejet (approved):
As discussed in slack, I'm fine with this. Could also be 350kb

PR #557, opened 2026-06-03
Fix our CI workflow on Windows

joanise:
### PR Goal? 

Fix test-for-windows

Was broken due to recent incompatibility between zip handlers and the latest node.

The latest cypress release https://docs.cypress.io/app/references/changelog#15-16-0 reports a fix for the problem as the first entry of the Bugfixes list:
> Fixed an issue on Node 24.16.0+ and Node 26.1.0+ where installing Cypress could silently extract only the first file from the binary archive, causing the test runner to fail to launch with a "Cypress binary is missing" error. Addresses [#33891](https://github.com/cypress-io/cypress/issues/33891). Addressed in [#33887](https://github.com/cypress-io/cypress/pull/33887).

At the same time, I'm re-applying #555 since it's no longer breaking CI. I think that was in fact unrelated, cypress was most likely the problem from the beginning.

### Fixes? 

broken Windows CI workflow

### Feedback sought? 

sanity check

### Priority? 

medium high

### How to test? 

look at the CI results for "Run tests on Windows"

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,7 +63,7 @@
         "@typescript-eslint/eslint-plugin": "^8.40.0",
         "@typescript-eslint/parser": "^8.40.0",
         "commitlint": "^20.5.0",
-        "cypress": "^15",
+        "cypress": "^15.16.0",
         "find-cache-dir": "^5.0.0",
         "husky": "^9.1.7",
         "jasmine-core": "^5.8.0",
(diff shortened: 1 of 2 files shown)


deltork (approved):
Simple fix 🤣!

joanise:
> Simple fix 🤣!

So much work to find it! 😅

PR #558, opened 2026-06-03
build: approve scripts we rely on at install time, in prep for npm 12

joanise:
### PR Goal? 

With npm 12, scripts will not run unless they are explicitly approved. 

See 
 - https://docs.npmjs.com/cli/v11/commands/npm-approve-scripts
 - https://ishu.dev/post/npm-install-scripts-opt-in-rfc-prepare-now-2026-05-20

### Fixes? 

Future proof our installation procedure, although I'm sure it's going to cause trouble when we bump versions in the package lock and forget to bump script approval. But this is where npm is heading, and that's going to make things safer, so we have to adapt to it.

### Feedback sought? 

Is it OK to blanket all the scripts we run now just like this? I'm figuring at this point we use these and things are OK. These permissions are pinned, so I'm figuring if a supply chain attack compromises something we use, it won't be able to run and we'll get a chance to audit before we update the allow list.

### Priority? 

low

### Tests added? 

n/a

### How to test? 

When you run `npm install` on main with a recent version of npm, you'll see this warning:
> npm warn allow-scripts 18 packages have install scripts not yet covered by allowScripts:

with this PR, that warning is gone.

You can see it in CI on main vs this PR's branch.

### Confidence? 

medium

### Version change? 

no

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,5 +86,18 @@
   },
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "allowScripts": {
+    "esbuild@0.25.5": true,
+    "esbuild@0.25.9": true,
+    "esbuild@0.27.7": true,
+    "nx@21.6.10": true,
+    "nx@21.6.11": true,
+    "nx@22.6.5": true,
+    "@parcel/watcher@2.5.6": true,
+    "cypress@15.16.0": true,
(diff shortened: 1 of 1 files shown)


joanise:
This will be a nuisance, we don't want to enable this before we have to.

PR #566, opened 2026-06-19
chore: npm audit fix

joanise:


joanise:
This is hopeless. No point trying to bump stuff while still on ng 20, we can try again any necessary remaining updates after #570 is merged.

PR #570, opened 2026-06-22
Bump Angular to version 21 and nx to latest

joanise:
### PR Goal? 

Bump Angular so that we can fix our many outstanding CVEs
 - angular is now on version 21
 - nx is now on version 23

### Fixes? 

Hopefully many CVEs...

### Feedback sought? 

sanity checking, and making sure my code changes make sense

### Priority? 

normal

### Tests added? 

actually, I had to remove the karma test run, because it's not compatible with ng21, but the end-to-end tests code the app well, and I tested it by hand too.

### How to test? 

play with the PR preview. I cannot see any differences myself, it would be good to confirm that

### Confidence? 

pretty good

### Version change? 

This will warrant a patch bump for sure.


joanise:
Just discovered: `provideZoneChangeDetection()` really has to be provided, otherwise our microphone buttons don't work properly anymore.

sergeleger (approved):
Site seems to be OK, I also didn't notice anything broken.

PR #571, opened 2026-06-23
Remove unused deprecated dependencies

joanise:
### PR Goal? 

When we run `npm install`, a number of deprecations are reported. Some of them we still use, but some we don't anymore. This PR removed unused deprecated dependencies.

### Feedback sought? 

Sanity checking

### Priority? 

normal

### Tests added? 

Already tested in CI

### How to test? 

See that CI passes.

### Confidence? 

high

### Version change? 

no


sergeleger (approved):
Most of this looks familiar with my WW angular upgrades. Looks good.

PR #577, opened 2026-06-25
chore: bump version to 1.6.4

joanise:
### PR Goal? 

Get ready to publish Studio-Web and Web-Component 1.6.4

### Feedback sought? 

Do you agree that we ready to publish 1.6.4 now?

### Confidence? 

high -- I've completed the angular bump and reviewed all outstanding CVE reports and I think it's a good idea to release now

### Version change? 

yup

PR #579, opened 2026-06-26
fix: make pause at end of page work with any translation

### LINKED ISSUE #578: [Web Component]: translation lines break "pause at end of page" setting
I was recently informed about this bug and have reproduced it below. If translation lines are present, and the user selects "Pause at the end of page", that functionality is broken. Without translation lines, the pausing happens, but with translation lines, the readalong plays through as normally. It doesn't seem to matter if the translations are displayed or not.

Once we find the fix, we should add this particular configuration to our e2e testing.

Working: [title-20260625163124.html](https://github.com/user-attachments/files/29360932/title-20260625163124.html)
Broken: [title-20260625163058.html](https://github.com/user-attachments/files/29360933/title-20260625163058.html)

joanise:
### PR Goal? 

We were excluding sentences that had both class .translation and class .sentence__translation from the list of sentences to figure how which is the last sentence in a page, but we should exclude any sentence with either of those classes, or with do-no-align=true.

Depending on how translations are added to a read-along, we don't always set both classes, so let's filter out anything that might be a translation when we are playing in the web-component.

### Fixes? 

Fixes #578

### Feedback sought? 

sanity checking

### Priority? 

high

### Tests added? 

no

### How to test? 

 - load the 058 example from #578 in the editor
 - click gear menu and select "pause at the end of each page"
 - click play
 - see the pause work again

### Confidence? 

high

### Version change? 

We'll want to release this quickly since a user stumbled on it. Let's merge this, then #577, and then let's publish 1.6.4.

### DIFF
--- a/packages/web-component/src/components/read-along-component/read-along.tsx
+++ b/packages/web-component/src/components/read-along-component/read-along.tsx
@@ -1125,7 +1125,8 @@ export class ReadAlongComponent {
             const sentences = paragraphs[
               paragraphs.length - 1
             ].querySelectorAll(
-              "s:not(.translation), s:not(.sentence__translation)",
+              // chain the not statements so they are ANDed: exclude sentences matching any of these criteria
+              's:not(.translation):not(.sentence__translation):not([do-not-align="true"])',
             ); //get non-translation sentences in the last paragraph
             const word =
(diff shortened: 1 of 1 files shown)


sergeleger (approved):
Bug is fixed, pauses at the end of the first page.

PR #582, opened 2026-07-03
fix(tests): fix spec tests after angular v21 upgrade

roedoejet:
I decided to fix this first since I wanted to run spec tests on the new code too. `npx nx test:once studio-web` now works.


joanise:
Urgh, so annoyingly easy. I just reverted the commit that disabled spec tests in CI, and and am ready to approve this PR once CI passes.

joanise (approved):

PR #583, opened 2026-07-03
Dev.ap/text overlay

roedoejet:
### PR Goal? 

The goal of this PR is to see how far we can get with a text overlay in solving some outstanding issues with the studio.

Namely:

1. Localized g2p error handling
2. Better paragraph/page boundary highlighting
3. Ability to add do-not-align spans

### Feedback sought? 

Functionality mostly - I've only implemented the first 2 (and haven't done DNA spans yet). It's finnicky since we're relating word indices from tokenized XML back to plain text and the highlighting has to persist even when the tokenization changes (e.g. as a user is fixing the text). I think this would be easier with a rich text editor since we could update the structured text and then map it all to XML when we upload to the API, but it seems possible this way, without the bloat, but maybe with some extra cognitive overhead.

### Priority? 

medium

### Tests added? 

Lots of spec tests for various scenarios

### How to test? 

`npx nx test:once studio-web`

### Confidence? 

medium. I got a lot of help from Claude Sonnet ( after planning it all out with @joanise) and some of the overlay css and alignment logic I'm just "trusting" with the help of spec tests and my own personal use of the site.

### TODO

- [ ] We can change the CSS for page/paragraph boundaries to be a little less loud. maybe also allow us to toggle it off
- [ ] I think we can remove the existing info button for page/paragraph boundaries and update it with add paragraph/page buttons
- [ ] I think the gutter icons for page/paragraph can be interactive (click to change/remove)
- [ ] We'll have to update the tour accordingly
- [ ] Add do-not-align segment UX
- [ ] Change and/or translate new toast message asking user to look at underlined g2p errored words

### DIFF
--- a/packages/studio-web/src/app/shared/shared.module.ts
+++ b/packages/studio-web/src/app/shared/shared.module.ts
@@ -2,13 +2,22 @@ import { NgModule } from "@angular/core";
 import { CommonModule } from "@angular/common";
 import { DownloadComponent } from "./download/download.component";
 import { TextareaOverlayComponent } from "./textarea-overlay/textarea-overlay.component";
+import { TextareaGutterComponent } from "./textarea-overlay/textarea-gutter.component";
 import { BrowserModule } from "@angular/platform-browser";
 import { MaterialModule } from "../material.module";
 import { FormsModule } from "@angular/forms";
 
 @NgModule({
-  declarations: [DownloadComponent, TextareaOverlayComponent],
+  declarations: [
+    DownloadComponent,
+    TextareaOverlayComponent,
+    TextareaGutterComponent,
+  ],
   imports: [BrowserModule, MaterialModule, FormsModule, CommonModule],
-  exports: [DownloadComponent, TextareaOverlayComponent],
+  exports: [
+    DownloadComponent,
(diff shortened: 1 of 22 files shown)


joanise:
Some thoughts based on a first pass of testing
 - I love the underline for g2p errors. That's clear and effective, just what we want.
 - Agreed, the para and page styling need to be more muted. Red says error to me, and that's not what we want. Maybe grey for a paragraph and blue for a page?
 - the paragraph and page markers show up too soon: if I type: 
   - text + enter + text: I get a P marker as soon as I hit enter before I start writing the second sentence. But the result has not paragraph break because I entered text.
   - text + enter + enter + text: I get a page marker displayed before I start typing the second text
 - instead, the P marker should show only only after typing text + enter + enter, and the page marker only after text + 3x enter. Maybe the simplest way to get there is to ignore the last document-final blank line, if any.
 - tested on my phone, and it's really unstable, the cursor is not shown where it is logically located once there is a paragraph break.

roedoejet:
Closing in favour of the TipTap approach: https://github.com/ReadAlongs/Studio-Web/pull/586 as discussed at the meeting today

PR #588, opened 2026-07-09
Fix broken Cypress CI workflows on both Windows and Linux

joanise:
### PR Goal? 

Fix running Cypress tests in CI.

On Linux, that mysteriously starting needed an explicit `npx cypress install`, not sure why that stopped happening implicitly.

On Windows, things were just a lot more broken, so I fell back to hand writing our steps to run Cypress like we do interactively, and really, it's much simpler, so maybe we should do the same on Linux too...

Bumping the version of Cypress because I first hoped that would fix the broken workflows. It did not, but we might as well keep the update.

### Fixes? 

Broken CI

### Feedback sought? 

sanity checking

### Priority? 

high -- CI is currently broken on main

### How to test? 

See CI results

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/.github/workflows/end-to-end-tests.yml
+++ b/.github/workflows/end-to-end-tests.yml
@@ -22,7 +22,7 @@ jobs:
           node-version: 24
       - run: npm install -g npm@latest
       - name: Install everything
-        run: npm ci
+        run: npm ci && npx cypress install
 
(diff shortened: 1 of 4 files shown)


roedoejet (approved):
lgtm!

PR #595, opened 2026-07-22
Dev.ej/bump nxext stencil

joanise:
### PR Goal? 

When bumping nx and ng, it seems I forgot to bump @nxext/stencil at the same time. This PR
 - bumps nxext/stencil to 23, in line with the rest, allowing several CVE patches to be compatible now
 - updates package-lock accordingly
 - `npm audit fix` to fix whatever else can be fixed automatically
 - fixed `stencil.config.js` to adapt to 23

### Fixes? 

inability to apply CVE updates due to locked dependency in nxext/stencil 21

### Feedback sought? 

just sanity checking, really.

I was able to confirm that with my adjustments to `stencil.config.js` are such that when we build, we get a identical `dist/packages/web-component` and bundles as before (modulo the contents of `stencil.config.js` itself).

### Priority? 

medium, as this will feed into fixing CVEs

### Tests added? 

n/a

### How to test? 

CI pretty much does it all, but to specifically evaluate the impact on building and bundling, I did `npx nx bundle web-component` while on `origin/main`, saved a copy if `dist/`, did it again on this branch and diffed the two structures. In fact, that was the process I followed to correctly configure all those `dir` fields in `stencil.config.js`: I adjusted them until the contents of `dist/` were identical before and after.

### Confidence? 

high, actually, given the dist and bundles are back to being identical.

### Version change? 

no

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,7 +50,7 @@
         "@nx/eslint": "^23.0.0",
         "@nx/eslint-plugin": "^23.0.0",
         "@nx/jest": "^23.0.0",
-        "@nxext/stencil": "^21.0.0",
+        "@nxext/stencil": "^23.0.0",
         "@playwright/test": "^1.60.0",
         "@stencil/angular-output-target": "^0.10.2",
         "@stencil/core": "^4.35.3",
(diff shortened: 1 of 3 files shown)


sergeleger (approved):
Seems fine to me.

